PR #5, opened 2015-02-19
Adds depends_on resource

athompson-r7:


lib/convection/model/template/resource.rb
```diff
@@ -51,10 +53,12 @@ def reference
...
 
         def render
-          {
+          resource = {
             'Type' => type,
-            'Properties' => properties
-          }.tap do |resource|
+            'Properties' => properties,
+          }
+          resource.merge!({'DependsOn' => depends_on}) unless depends_on.empty?
```
jmanero-r7: This should just go inside the tap codeplock

lib/convection/model/template/resource.rb
```diff
@@ -34,6 +35,7 @@ def initialize(name, template)
 
           @type = ''
           @properties = {}
+          @depends_on = ''
```
jmanero-r7: The docs say that this can actually be an array. Could it be
```ruby
def depends(resource)
  @depends_on << resource
end
```
instead?

jmanero-r7:
Looks good :+1:

PR #8, opened 2015-02-24
added iam user

rhodgman-r7:
added template resource representing iam user
updated iam policy template resource to handle users (and groups)


example/iam_user.rb
```diff
@@ -0,0 +1,31 @@
...
+        :Action => ['s3:GetObject'],
+        :Resource => ['arn:aws:s3:::some.bucket.name.here/*']
+      }]
+    )
+  end
+
+  iam_role 'NewUser' do
+    path fn_ref(:Path)
+  end
+end
```
athompson-r7: Should this be a call to `iam_user`, because you are calling `user` in the policy?

rhodgman-r7: Yup, nice catch, I'll fix that up soon.

rhodgman-r7:
I updated the example, this should be good to know now. Also, at some point we may want to go through and make these examples a little more flushed out, using some more of the available properties and methods.

jmanero-r7:
:+1:

PR #11, opened 2015-02-25
Adds support for Arrays in the property helper

athompson-r7:
To use this with an array property do the following:

    property :vpc_security_groups, 'VPCSecurityGroups', :array

This will initialize `@properties['VPCSecurityGroups']` to an empty Array,
then adds elements to `@properties['VPCSecurityGroups']` as
`vpc_security_groups` is called within the dsl.


jmanero-r7:
:+1:

PR #21, opened 2015-06-08
Stop. Hamer Time.

jmanero-r7:
* Big factor down of noisy DSL boilerplate in resources
* Improved namespace management. resource DSL is now available within other resources (e.g. define ec2_instances inside of a VPC resource's closure)
* Add per-resource-property state tracking. Load resources and templates from the CF API at startup and populate a 'current_state' value in each PropertyInstance.


jmanero-r7:
This branch is deprecated. Branch v1.0 has been rebased from this branch onto master. See README.md in [hamer-time](/rapid7/convection/tree/hamer-time) for details

PR #24, opened 2015-07-20
CORE-1058 Generate warning text in template for NULL stack/key refere…

simonirwin-r7:
…nces

This change will prevent the following JSON:
   "Type": "AWS::EC2::Instance",
      "Properties": {
        "SecurityGroupIds": [
          null
        ],

Instead it will generate the folllowing:
      "Type": "AWS::EC2::Instance",
      "Properties": {
        "SecurityGroupIds": [
          "** Warning: Could not resolve value [missing-resource] for stack [unknown-security-groups] **"
        ],


lib/convection/control/stack.rb
```diff
@@ -105,7 +105,7 @@ def []=(key, value)
       end
 
       def get(*args)
-        @attributes.get(*args)
+        @attributes.get(*args, "** Warning: Could not resolve value [#{args[1]}] for stack [#{args[0]}] **")
```
athompson-r7: I think we should throw an exception here rather than adding some bogus text that will fail to evaluate in the CloudFormation controller.

simonirwin-r7: I think we could track something like stack.error_count - we'd bump this integer each time there's a null reference to a stack (or key).  We would print something sensible so that 'convection print' completes with some useful JSON for testing/debugging, but the operation would throw an exception (because stack.error_count != 0) and the template would NOT be passed to CF.

what do you think?

athompson-r7: [This](https://github.com/rapid7/convection/blob/master/lib/convection/model/attributes.rb#L43-L45) is really where the issue is, though, the question is do we want it to fail during template render or during apply/diff stage.  @jmanero-r7 Care to weigh in here?

jmanero-r7: I think we should throw an exception. That should surface a stack trace, and give the user an idea of where in the template the null value is getting generated:

```ruby
@attributes.get(*args) { fail, SomeException, 'Attempted to reference an undefined attribute!' }

rvanburen-r7:
Bundle exec convection dif still works even if some of the stack is missing

╭─rvanburen@rapid7.com ~/Git/razor-convection/clouds/us-east-1/razor-test-2 ‹2.2.2› ‹master*›
╰─➤  bundle exec convection dif 
     compare  Compare local state of stack vpc (razor-test-2-vpc) with remote template
      create  Stack razor-test-2-vpc is not yet defined
         . . .
     compare  Compare local state of stack manage (razor-test-2-manage) with remote template
      create  Stack razor-test-2-manage is not yet defined

Bundle exec convection print #{ template file} will now throw an error if the stack is not defigned properly

╭─rvanburen@rapid7.com ~/Git/razor-convection/clouds/us-east-1/razor-test-2 ‹2.2.2› ‹master*›
╰─➤  bundle exec convection print jumphost 
/Users/rvanburen/Git/razor-convection/templates/ec2/jump_host.rb:30:in `block (2 levels) in <module:EC2>': Attempted to reference an undefined attribute! (ArgumentError)
	from /Users/rvanburen/Git/convection/lib/convection/model/template.rb:21:in `instance_exec'
	from /Users/rvanburen/Git/convection/lib/convection/model/template.rb:21:in `block in 
         . . .
  (Rest of the stack trace)

rvanburen-r7:
Closing this pull request to open it against current development branch [v0.2]

PR #26, opened 2015-07-24
IAM Profile support to EC2 instances

athompson-r7:
Chery picking into this branch from master...

Also fixes hash format for newer Ruby version


athompson-r7:
Wrong branch.

Issue #31, opened 2015-08-17
AWS rate limiting


athompson-r7:
As the number of stacks in a Cloudfile groups we are hitting the AWS rate limiting.  We should detect this and try to slow down a bit.

* Rate exceeded
    /Users/athompson/.chefdk/gem/ruby/2.1.0/gems/aws-sdk-core-2.1.14/lib/seahorse/client/plugins/raise_response_errors.rb:15:in `call'
    /Users/athompson/.chefdk/gem/ruby/2.1.0/gems/aws-sdk-core-2.1.14/lib/aws-sdk-core/plugins/param_converter.rb:21:in `call'
    /Users/athompson/.chefdk/gem/ruby/2.1.0/gems/aws-sdk-core-2.1.14/lib/aws-sdk-core/plugins/response_paging.rb:26:in `call'
    /Users/athompson/.chefdk/gem/ruby/2.1.0/gems/aws-sdk-core-2.1.14/lib/seahorse/client/plugins/response_target.rb:21:in `call'
    /Users/athompson/.chefdk/gem/ruby/2.1.0/gems/aws-sdk-core-2.1.14/lib/seahorse/client/request.rb:70:in `send_request'
    /Users/athompson/.chefdk/gem/ruby/2.1.0/gems/aws-sdk-core-2.1.14/lib/seahorse/client/base.rb:207:in `block (2 levels) in define_operation_methods'
    /Users/athompson/code/convection/lib/convection/control/stack.rb:254:in `get_status'
    /Users/athompson/code/convection/lib/convection/control/stack.rb:182:in `apply'
    /Users/athompson/code/convection/lib/convection/control/cloud.rb:30:in `block in converge'
    /Users/athompson/code/convection/lib/convection/control/cloud.rb:28:in `each'
    /Users/athompson/code/convection/lib/convection/control/cloud.rb:28:in `converge'
    /Users/athompson/code/convection/bin/convection:20:in `converge'
    /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
    /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
    /opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
[...]

PR #33, opened 2015-08-18
Adds some basic rate limiting

athompson-r7:


lib/convection/control/stack.rb
```diff
@@ -201,7 +202,17 @@ def apply(&block)
 
         watch(&block) if block # Block execution on stack status
       rescue Aws::Errors::ServiceError => e
-        @errors << e
+        if e.message == 'Rate exceeded'
+          operation = context.operation_name
```
jmanero-r7: Where's `context` coming from?

athompson-r7: Ooops, I must have trimmed that when cleaning up.

lib/convection/control/stack.rb
```diff
@@ -85,6 +85,7 @@ def initialize(name, template, options = {})
         @resources = {}
         @current_template = {}
         @last_event_seen = nil
+        @retries = 0
```
athompson-r7: I'm not exactly sure if I needed to reset `@retries` within each method that was going to use this or if we get a new instance of this class each time.

jmanero-r7: As far as I remember here, the Stack is only instantiated at startup when the Cloudfile is parsed.

athompson-r7:
@jmanero-r7 simplified this a bit.

athompson-r7:
Fixes #31

PR #34, opened 2015-08-18
Diff new stack

athompson-r7:
This allows us to diff a new stack that hasn't yet been converged.  This
is useful to do a last verification before converging.


jmanero-r7:
:+1:

PR #35, opened 2015-08-19
Add an RDS option to allow encryption.

alee-r7:
For Nexpose, we want our RDS instances to have encryption enabled.

I created a PR on top of hamer-time. Seems razor-convection is dependent on this branch from github.

### DIFF
--- a/lib/convection/model/template/resource/aws_rds_db_instance.rb
+++ b/lib/convection/model/template/resource/aws_rds_db_instance.rb
@@ -30,6 +30,7 @@ module Convection
           property :option_group, 'OptionGroupName'
 
           property :storage_encrypted, 'StorageEncrypted'
+          property :kms_key_id, 'KmsKeyId'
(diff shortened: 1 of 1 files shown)


athompson-r7:
LGTM, though we likely want to add `KmsKeyId` too.  If we don't it will use the default key.

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-kmskeyid

alee-r7:
Good idea. Do you want me to add that as part of this PR?

athompson-r7:
Yeah if you can.  Once it is encrypted it cannot be unencrypted, so we have to get it right.

alee-r7:
Ok, will do. BTW: You can still access the data using default key and the mysql username/password. So if you are worried about accessing the data it shouldn't be an issue.

PR #36, opened 2015-08-28
CORE-1136: Add template Validation

aburke-r7:
**Validation Added for:**
- template body bytesize
- number of characters in description
- number of resources
- number of parameters
- number of characters in parameter name
- number of mappings
- number of outputs
- number of characters in each output name

NEW
- number of characters in each mapping name
- number of mapping attributes
- number of characters in each mapping attribute name
- parameter value bytesize
- number of characters in each resource name

[**In line with AWS limits**] (http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html)


lib/convection/model/template.rb
```diff
@@ -183,8 +183,75 @@ def diff(other, stack_ = nil)
       end
 
       def to_json(stack_ = nil, pretty = false)
-        return JSON.generate(render(stack_)) unless pretty
-        JSON.pretty_generate(render(stack_))
+        stack = render(stack_)
```
jmanero-r7: `stack` is going to shadow the instance method

lib/convection/model/template.rb
```diff
@@ -183,8 +183,75 @@ def diff(other, stack_ = nil)
...
+
+        #bytesize
+        json = JSON.generate(stack)
+        template_bytesize = json.bytesize
+        cf_max_bytesize = 51200
```
jmanero-r7: This should be defined as a constant in the scope of the class instead of a variable in the method

lib/convection/model/template.rb
```diff
@@ -183,8 +183,75 @@ def diff(other, stack_ = nil)
...
+        if template_bytesize > cf_max_bytesize
+          Kernel.abort("Error: Excessive Template Size (#{template_bytesize}) Max= #{cf_max_bytesize}")
```
jmanero-r7: We should raise Exceptions for failed validation instead of just printing text

athompson-r7:
While this will probably work, I think I will make more sense for this to be in unit/spec test form right now.  This way we can run tests on each commit and quickly determine if there might be issues.  This paired with ongoing talks of rewriting some of the internals.

Below is an example of how you can, given a Cloudfile, render the json for each stack.  After the json is rendered multiple tests can be run on that json.

```ruby
require 'convection/control/cloud'
cloud = ::Convection::Model::Cloudfile.new('./clouds/eu-central-1/razor-dev-0/Cloudfile')
cloud.deck.each do |stack|
  rendered_json = stack.template.render
  # do something
end
```

lib/convection/model/template.rb
aburke-r7: @jmanero-r7 Are you happy with it also printing the stack trace?

test/test_helper.rb
```diff
@@ -5,7 +5,14 @@
   add_group 'DSL', 'lib/convection/dsl'
 end
 
+
 gem 'minitest'
 require 'minitest/autorun'
 require 'minitest/spec'
 require_relative '../lib/convection'
+
+module Minitest::Assertions
```
aburke-r7: (Adding support for assert_nothing_raised)

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
+        resources=rendered_stack["Resources"]
+        number_of_resources = resources.count
+        if number_of_resources > CF_MAX_RESOURCES
+          raise ArgumentError, "Error: Excessive Number of Resources (#{number_of_resources}) Max= #{CF_MAX_RESOURCES}"
+        end
+        largest_resource_name = resources.keys.max
+        if largest_resource_name == nil
+          largest_resource_name = ""
+        end
+        resource_name_characters = largest_resource_name.length
```
simonirwin-r7: Shouldn't use a default of 'nil' for the rendered_stack these validation_XXX methods

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
+        if  resource_name_characters> CF_MAX_RESOURCE_NAME
+          raise ArgumentError, "Error: Resource Name #{largest_resource_name} has too many characters (#{resource_name_characters}) Max=#{CF_MAX_RESOURCE_NAME} "
+        end
+      end
+
+      def validate_mappings(rendered_stack = nil)
```
simonirwin-r7: Shouldn't use a default of 'nil' for the rendered_stack these validation_XXX methods

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
+
+      end
+
+      def validate_resources(rendered_stack = nil)
+        resources=rendered_stack["Resources"]
+        number_of_resources = resources.count
+        if number_of_resources > CF_MAX_RESOURCES
+          raise ArgumentError, "Error: Excessive Number of Resources (#{number_of_resources}) Max= #{CF_MAX_RESOURCES}"
+        end
+        largest_resource_name = resources.keys.max
```
simonirwin-r7: I think you're trying to find the longest resource name in the array?  I don't think that 'max' will do it.

see: http://apidock.com/ruby/Enumerable/max_by

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
 
       def to_json(stack_ = nil, pretty = false)
+        rendered_stack = render(stack_)
+        validate(rendered_stack)
+
         return JSON.generate(render(stack_)) unless pretty
         JSON.pretty_generate(render(stack_))
       end
+
+      def validate(rendered_stack = nil)
```
simonirwin-r7: This looks a lot better!

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
+      end
+
+      def validate_resources(rendered_stack = nil)
+        resources=rendered_stack["Resources"]
+        number_of_resources = resources.count
+        if number_of_resources > CF_MAX_RESOURCES
+          raise ArgumentError, "Error: Excessive Number of Resources (#{number_of_resources}) Max= #{CF_MAX_RESOURCES}"
+        end
+        largest_resource_name = resources.keys.max
+        if largest_resource_name == nil
```
simonirwin-r7: You *could* condense this to:

largest_resource_name ||= ''

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
+
+      def validate_mappings(rendered_stack = nil)
+        mappings = rendered_stack ['Mappings']
+        number_of_mappings = mappings.count
+        if number_of_mappings > CF_MAX_MAPPINGS
+          raise ArgumentError, "Error: Excessive Number of Mappings (#{number_of_mappings}) Max= #{CF_MAX_MAPPINGS}"
+        end
+        mappings.each do |key,value|
+          if value.count > CF_MAX_MAPPING_ATTRIBUTES
+            raise ArgumentError, "Error: Excessive Number of Mapping Attributes"
```
simonirwin-r7: Mention the max in the error message?

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
+          end
+        end
+        mappings.keys.each do |key|
+          if key.length > CF_MAX_MAPPING_NAME
+            raise ArgumentError, "Error: Excessive Mapping Name #{key}"
```
simonirwin-r7: Mention the max in the error message?

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
+          largest_parameter_name = ""
+        end
+        parameter_name_characters = largest_parameter_name.length
+        if parameter_name_characters > CF_MAX_PARAMETER_NAME_CHARACTERS
+          raise ArgumentError, "Error: Parameter Name #{largest_parameter_name} has too many characters (#{parameter_name_characters}) Max= #{CF_MAX_PARAMETER_NAME_CHARACTERS}"
+        end
+        parameters.values.each do |value|
+          parameter_in_json = JSON.generate(value)
+          if parameter_in_json.bytesize > CF_MAX_PARAMETER_VALUE_BYTESIZE
+            raise ArgumentError, "Error: Excessive Parameter Size #{value}"
```
simonirwin-r7: Mention the max?

lib/convection/model/template.rb
```diff
@@ -11,6 +11,20 @@ module DSL
     module Template
       include DSL::Helpers
 
+      CF_MAX_BYTESIZE = 51200
```
erran-r7: For numbers greater than 999 use the underscore character. `51_200` it helps legibility.

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
       end
 
       def to_json(stack_ = nil, pretty = false)
+        rendered_stack = render(stack_)
+        validate(rendered_stack)
+
         return JSON.generate(render(stack_)) unless pretty
```
erran-r7: Instead of rendering the stack multiple lines re-use the `rendered_stack` variable here and on the line below.

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
+      def validate_bytesize(rendered_stack = nil)
+        json = JSON.generate(rendered_stack)
+        template_bytesize = json.bytesize
+        if template_bytesize > CF_MAX_BYTESIZE
+          raise ArgumentError, "Error: Excessive Template Size (#{template_bytesize}) Max= #{CF_MAX_BYTESIZE}"
+        end
+      end
+
+
+
```
simonirwin-r7: You could take this a step further and have something like:

```
  def check_value_is_within_limits(val, limit, description)
    raise ArgumentError, "Error: #{description} is #{val}; limit is #{limit}" if val > limit
  end
```

  You would call it as follows:

```
  check_value_is_within_limits(json.bytesize, CF_MAX_BYTESIZE, 'template size')
```

This should look a bit cleaner, and maybe get rid of a few local variables, too.

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
+        validate_mappings(rendered_stack)
+        validate_parameters(rendered_stack)
+        validate_outputs(rendered_stack)
+        validate_description(rendered_stack)
+        validate_bytesize(rendered_stack)
+
+      end
+
+      def validate_resources(rendered_stack = nil)
+        resources=rendered_stack["Resources"]
```
erran-r7: space on either side of the `=` operator.

lib/convection/model/template.rb
erran-r7: That or add `.to_s` to the end of call to `resources.keys.max`.

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
+        number_of_resources = resources.count
+        if number_of_resources > CF_MAX_RESOURCES
+          raise ArgumentError, "Error: Excessive Number of Resources (#{number_of_resources}) Max= #{CF_MAX_RESOURCES}"
+        end
+        largest_resource_name = resources.keys.max
+        if largest_resource_name == nil
+          largest_resource_name = ""
+        end
+        resource_name_characters = largest_resource_name.length
+        if  resource_name_characters> CF_MAX_RESOURCE_NAME
```
erran-r7: Spaces on both sides of the `>` operator.

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
+
+      def validate_parameters(rendered_stack = nil)
+        parameters= rendered_stack['Parameters']
+        number_of_parameters = parameters.count
+        if number_of_parameters > CF_MAX_PARAMETERS
+          raise ArgumentError, "Error: Excessive Number of Parameters (#{number_of_parameters}) Max= #{CF_MAX_PARAMETERS}"
+        end
+        largest_parameter_name = parameters.keys.max
+        if largest_parameter_name == nil
+          largest_parameter_name = ""
```
erran-r7: See @simonirwin-r7's previous comment about using `||=` or use `.to_s` on max in which case nil will cause the result to be empty string.

lib/convection/model/template.rb
```diff
@@ -183,9 +197,122 @@ def diff(other, stack_ = nil)
...
+
+      def validate_outputs(rendered_stack = nil)
+        outputs = rendered_stack['Outputs']
+        number_of_outputs = outputs.count
+        if number_of_outputs > CF_MAX_OUTPUTS
+          raise ArgumentError, "Error: Excessive Number of Outputs (#{number_of_outputs}) Max= #{CF_MAX_OUTPUTS}"
+        end
+        largest_output_name = outputs.keys.max
+        if largest_output_name == nil
+          largest_output_name = ""
```
erran-r7: > See @simonirwin-r7's previous comment about using ||= or use .to_s on max in which case nil will cause the result to be empty string.

lib/convection/model/template.rb
erran-r7: Extra whitespace.

test/convection/model/test_validation.rb
```diff
@@ -0,0 +1,184 @@
...
+          property 'ImageId', 'ami-76e27e1e'
+          property 'KeyName', 'test'
+          property 'SecurityGroupIds', ['sg-dd733c41', 'sg-dd738df3']
+          property 'Tags', [{
+            'Key' => 'Name',
+            'Value' => 'test-1'
+            }]
+          end
+        end
+      10.times do |count|
```
erran-r7: It looks like you could reduce the number of calls to `10.times`.

test/convection/model/test_validation.rb
```diff
@@ -0,0 +1,184 @@
...
+        end
+      100.times do |count|
+        mapping "Mapping_#{count}" do
+          item 'us-east-1', 'test', 'cf-test-keys'
+          item 'us-west-1', 'test', 'cf-test-keys'
+        end
+      end
+    end
+    @excessive_description = ::Convection.template do
+      description "0" * 1025
```
erran-r7: Instead of hardcoding this why not use the constant that defines the excessive length `+ 1`?

test/convection/model/test_validation.rb
```diff
@@ -0,0 +1,184 @@
...
+        description 'An Important Attribute'
+        value get_att('Resource', 'Attribute')
+      end
+    end
+  end
+
+  def test_within_limits
+    assert_nothing_raised(ArgumentError) {@within_limits.to_json}
+  end
+  def test_bytesize
```
erran-r7: Newline between `def`/`end`.

test/convection/model/test_validation.rb
```diff
@@ -0,0 +1,184 @@
+require 'test_helper'
+require 'json'
+require 'pp'
+
+class TestValidations < Minitest::Test
+
+  def setup
+    @within_limits = ::Convection.template do
```
erran-r7: `setup` probably isn't where you want to create all these templates. The method is run before every test method (methods named `test_*`). Move each of these declarations into their respective test.

test/convection/model/test_validation.rb
```diff
@@ -0,0 +1,184 @@
...
+          property 'AvailabilityZone', 'us-east-1a'
+          property 'ImageId', 'ami-76e27e1e'
+          property 'KeyName', 'test'
+          property 'SecurityGroupIds', ['sg-dd733c41', 'sg-dd738df3']
+          property 'Tags', [{
+            'Key' => 'Name',
+            'Value' => 'test-1'
+            }]
+          end
+        end
```
erran-r7: This `end`, the one above it, and the `}]`, are all intended two spaces too far.

test/convection/model/test_validation.rb
erran-r7: This will also make the tests **MUCH** slower.

test/convection/model/test_validation.rb
```diff
@@ -0,0 +1,184 @@
...
+      description 'Validations Test Template - Too Many Resources'
+      201.times do |count|
+        logs_log_group "Log_Group_#{count}" do
+          retention_in_days 365
+        end
+      end
+    end
+    @excessive_resource_name = ::Convection.template do
+      description 'Validations Test Template - Excessive Resource Name'
+        logs_log_group "0"*256  do
```
erran-r7: Use the excessive resource name constant here.

erran-r7: Also introduce an intention revealing variable name instead of having an at-first-glance-random call to `String#*`.

test/convection/model/test_validation.rb
```diff
@@ -0,0 +1,184 @@
...
+          retention_in_days 365
+        end
+      end
+    @excessive_mappings = ::Convection.template do
+      description 'Validations Test Template - Too Many Mappings'
+      101.times do |count|
```
erran-r7: Use a constant instead of `101` here.

test/convection/model/test_validation.rb
```diff
@@ -0,0 +1,184 @@
...
+        mapping "Mapping_#{count}" do
+          item 'us-east-1', 'test', 'cf-test-keys'
+          item 'us-west-1', 'test', 'cf-test-keys'
+        end
+      end
+    end
+    @excessive_mapping_attributes = ::Convection.template do
+      description 'Validations Test Template - Too Many Mapping Attributes'
+      mapping "Mapping_Example" do
+        31.times do |count|
```
erran-r7: Another [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)), use a constant instead.

test/convection/model/test_validation.rb
```diff
@@ -0,0 +1,184 @@
...
+          item "#{count}_1", "#{count}_2", "#{count}_3"
+        end
+      end
+    end
+    @excessive_mapping_name = ::Convection.template do
+      description 'Validations Test Template - Excessive Mapping Name'
+      mapping "0" *256 do
```
erran-r7: Same here, another magic number. Fix this and the occurrences below.

test/convection/model/test_validation.rb
```diff
@@ -0,0 +1,184 @@
...
+        item "1","2","3"
+      end
+    end
+    @excessive_mapping_attribute_names = ::Convection.template do
+      description 'Validations Test Template - Excessive Mapping Attribute Name'
+      mapping "Mapping_1" do
+        item "0" *256,"value","value"
```
erran-r7: Spaces after every comma.

test/convection/model/test_validation.rb
```diff
@@ -0,0 +1,184 @@
...
+        value get_att('Resource', 'Attribute')
+      end
+    end
+  end
+
+  def test_within_limits
+    assert_nothing_raised(ArgumentError) {@within_limits.to_json}
+  end
+  def test_bytesize
+    assert_raises(ArgumentError, "Error: Excessive Template Size") {JSON.generate(@excessive_bytesize.to_json).bytesize}
```
erran-r7: Spaces after `{` and before `}` for single line blocks. Note some of these blocks are pretty long. Consider using the `do`/`end` format instead.

test/test_helper.rb
```diff
@@ -5,7 +5,14 @@
   add_group 'DSL', 'lib/convection/dsl'
 end
 
+
```
erran-r7: Remove this added newline.

test/test_helper.rb
```diff
@@ -5,7 +5,14 @@
...
 gem 'minitest'
 require 'minitest/autorun'
 require 'minitest/spec'
 require_relative '../lib/convection'
+
+module Minitest::Assertions
+  def assert_nothing_raised(*)
```
erran-r7: This method literally only calls the block provided (ignoring all arguments). You should define this method so that it raises an assertion error if anything *was* raised.

erran-r7: Also, doesn't a `refute_raised` exist? Sounds like it'd be built-in.

test/convection/model/test_validation.rb
```diff
@@ -0,0 +1,184 @@
...
+      description 'Validations Test Template - Excessive Output Name'
+      output "0" * 256 do
+        description 'An Important Attribute'
+        value get_att('Resource', 'Attribute')
+      end
+    end
+  end
+
+  def test_within_limits
+    assert_nothing_raised(ArgumentError) {@within_limits.to_json}
```
erran-r7: Instead of testing that nothing is raised make a simple assertion that proves the object is valid.

```ruby
json = @within_limits.to_json
template = JSON.parse(json)
# assert that JSON matches what you expected.
```

test/test_helper.rb
erran-r7: I'd probably remove this and refactor the test(s) you use this in.

lib/convection/model/exceptions.rb
```diff
@@ -0,0 +1,17 @@
...
+ExcessiveParameterNameError = Class.new( StandardError)
+ExcessiveParameterBytesizeError = Class.new( StandardError)
+ExcessiveOutputsError = Class.new( StandardError)
+ExcessiveOutputNameError = Class.new( StandardError)
+ExcessiveDescriptionError = Class.new( StandardError)
+ExcessiveTemplateSizeError = Class.new( StandardError)
+
+def LimitExceededError(value, limit, errorClass)
+  raise errorClass, "Value #{value} exceeds Limit #{limit}"
+end
```
athompson-r7: Method names should be under_score_and_lower_case.  It looks like this method is a class being called wrong in the code.

https://github.com/rapid7/ruby-style-guide/blob/master/README.md#naming

athompson-r7: It also might make sense to subclass all of the specific limit exceptions with a more general limit exception so that we can rescue on general limit errors.

lib/convection/model/template.rb
```diff
@@ -1,6 +1,7 @@
 require_relative '../dsl/helpers'
 require_relative '../dsl/intrinsic_functions'
 require_relative './diff'
+require_relative 'exceptions.rb'
```
jmanero-r7: `require_relative './exceptions'

lib/convection/model/template.rb
```diff
@@ -183,8 +198,121 @@ def diff(other, stack_ = nil)
...
+        rendered_stack = render(stack_)
+        validate(rendered_stack)
+
+        return JSON.generate(rendered_stack) unless pretty
+        JSON.pretty_generate(rendered_stack)
+      end
+
+      def validate(rendered_stack = nil)
+        %w(resources mappings parameters outputs description bytesize).map do
+          |method| self.send("validate_#{method}", rendered_stack)
```
jmanero-r7: * Unnecessary use of `self`
* Put the argument list (`|...|`) on the same line as `do`/`{`

simonirwin-r7:
:+1:

PR #37, opened 2015-09-03
Add support for encryption at rest for RDS.

rhass-r7:
This adds support for encryption at rest for RDS instance types which support the feature.

On a related note, AWS::EC2::Volume needs to be created to be able to support this globally for all EC2 instances which support disk encryption.

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-ebs-volume.html


athompson-r7:
This was already added to the code base in #35.  I'm ok adding it to master like this.

On the subject of the `AWS::EC2::Volume` resource, if you want to tackle that be my guest.  We don't to separate volume attachments so we haven't had the need.

PR #38, opened 2015-09-08
Adds ability to configure the splay/sleep

athompson-r7:
This commit adds configurable splay time beteen each stack converge

To enable add the following to the Cloudfile:

```
splay 10
```


athompson-r7:
Wrong branch

PR #40, opened 2015-09-10
Core-1172 trust flow logs

rvanburen-r7:
Created trust_flow_logs that will set the correct trust permissions of flow log iam roles.


athompson-r7:
:+1:

simonirwin-r7:
:+1:

PR #41, opened 2015-09-10
Create .travis.yml

aburke-r7:


.travis.yml
```diff
@@ -0,0 +1,3 @@
+language: ruby
+rvm:
+  - 2.2.2
```
athompson-r7: Can you open up the range of rubies we test on?  Say 2.0+

Also if you add `sudo: false` it should build faster.

@jmanero-r7 care to weigh in?

aburke-r7: @athompson-r7 Addressed in the most recent commit

athompson-r7:
:+1:

PR #44, opened 2015-09-16
Add security group destination attribute

bturner-r7:
Adds a `destination` attribute for egress security groups.


simonirwin-r7:
:+1:

PR #46, opened 2015-09-17
Allow specifying network_interfaces in ec2_instances

bturner-r7:
This adds support for the [EC2 NetworkInterface Embedded Property Type](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-network-iface-embedded.html).  This allows you to declare network interfaces as part of an EC2 instance.  As far as I know, this is the only way to auto assign a public IP address when the VPC's auto-assign public IP is false.

For example, this supports:

```ruby
ec2_instance 'MyInstance' do
  image_id 'ami-12345678'
  instance_type 't2.micro'
  network_interface do
    associate_public_ip_address true
    security_group 'sg-12345678'
    subnet 'subnet-12345678'
  end
  network_interface do
    associate_public_ip_address false
    security_group 'sg-87654321'
    subnet 'subnet-12345678'
  end
end
```

If this is accepted, we could probably build a much improved DSL around [Resource Property Types](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-product-property-reference.html).


simonirwin-r7:
Interesting...are you actually looking to create network interfaces on 2 different subnets?  IIRC (and not sure how relevant it is to you) AWS doesn't support scaling an EC2 instance with more than one interface.

bturner-r7:
@simonirwin-r7, no I'm not looking to add two network interfaces. I just used that example to show how it combines individual calls into the list that `NetworkInterfaces` requires (and [sets](https://github.com/rapid7/convection/blob/add-network-interface/lib/convection/model/template/resource/aws_ec2_instance.rb#L29) the required `DeviceIndex` for you).

What I do need to do (in one of my other convection projects) is set the `AssociatePublicIpAddress` to true in a VPC that has auto-assign disabled.  This is the only way I've found to do this and the reason for this PR.

jmanero-r7:
:+1: Looks good

PR #47, opened 2015-09-17
Only build master and master-based pull requests

erran-r7:
cc @bturner-r7 @athompson-r7 @eciramella-r7 

---

Thoughts? This *should* fix the weird behaviour of building feature branches twice. We can add other branches to the `branches.only` array which accepts strings/regex. We could add a new line `- /^(release|staging)\/.*$/` in the YAML for example to build PRs/commits to branches named using the staging/release convention.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+branches:
+  only:
+    - master
 language: ruby
 rvm:
   - 2.0.0


erran-r7:
@aburke-r7 FYI in regards to the email @bturner-r7 sent earlier.

bturner-r7:
Thanks @erran-r7.  Actually I don't think this is an issue after all.  I got confused because the travis build number just happened to correspond to the pull request number and got confused.  We can probably close this - building all branches is fine for now.

erran-r7:
> Thanks @erran-r7. Actually I don't think this is an issue after all. I got confused because the travis build number just happened to correspond to the pull request number and got confused. We can probably close this - building all branches is fine for now.

Sounds good to me.

PR #51, opened 2015-09-24
Allow SQS policy to be defined like S3 policies.

alee-r7:
SQS policies are similar to S3 and SNS. No reason for SQS policies to deviate from SQS and S3

### DIFF
--- a/lib/convection/model/template/resource/aws_sqs_queue_policy.rb
+++ b/lib/convection/model/template/resource/aws_sqs_queue_policy.rb
@@ -12,7 +12,7 @@ module Convection
           extend Forwardable
 
           type 'AWS::SQS::QueuePolicy'
-          property :queues, 'Queues'
+          property :queue, 'Queues', :type => :list
           attr_reader :document
 
           def_delegators :@document, :allow, :id, :version, :statement
--- a/lib/convection/model/template/resource/aws_sqs_queue_policy.rb
+++ b/lib/convection/model/template/resource/aws_sqs_queue_policy.rb
@@ -1,3 +1,4 @@
+require 'forwardable'
 require_relative '../resource'
 
 module Convection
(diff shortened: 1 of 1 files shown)


lib/convection/model/template/resource/aws_sqs_queue_policy.rb
```diff
@@ -8,9 +9,25 @@ class Resource
         # AWS::SQS::QueuePolicy
         ##
         class SQSQueuePolicy < Resource
+          extend Forwardable
+
           type 'AWS::SQS::QueuePolicy'
-          property :queue, 'Queues', :type => :list
-          property :policy_document, 'PolicyDocument'
+          property :queues, 'Queues'
```
athompson-r7: Is there a reason for the change from `queue` to `queues`?  If is truly an array, the accumulator is the better approach.

    property :queue, 'Queues', :type => :list


Then in the template it would be this:

    queue 'somequeue'
    queue 'someotherqueue'

jmanero-r7: ^^ +1

`property :queue, 'Queues', :type => :list` is the correct usage

alee-r7: Yeah it is. Here is the CF json.
{
   "Type" : "AWS::SQS::QueuePolicy",
   "Properties" : {
      "PolicyDocument" : JSON,
      "Queues" : [ String, ... ]
   }
}

So just change this back to property :queue, 'Queues, :type => :list?

athompson-r7: Yes.

athompson-r7: For reference:
`:queue` - the accessor in ruby
`Queues` - the Cloudformation JSON key
`:type` - data, defaults to `:string`

athompson-r7:
Looks good.  Thanks for fixing this up!

PR #61, opened 2015-10-27
added lambda template

ndavis-r7:
[Cloudformation Documentation](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-role).

Example JSON:
```
"AMIIDLookup": {
  "Type": "AWS::Lambda::Function",
  "Properties": {
    "Handler": "index.handler",
    "Role": { "Fn::GetAtt" : ["LambdaExecutionRole", "Arn"] },
    "Code": {
      "S3Bucket": "lambda-functions",
      "S3Key": "amilookup.zip"
    },
    "Runtime": "nodejs",
    "Timeout": "25"
  }
}
```


lib/convection/model/template/resource/aws_lambda.rb
```diff
@@ -0,0 +1,22 @@
+require_relative '../resource'
```
athompson-r7: Can you rename this file to `aws_lambda_function.rb` to fit the current convention?

ndavis-r7: Can do

jmanero-r7:
:+1: This looks good. Needs to be rebased back up to v0.2 before merging.

ndavis-r7:
@jmanero-r7 I'm not very familiar with rebase. I did:
```
git checkout lambda-support
git rebase origin/v0.2
git push origin lambda-support
```

But then when I try to push onto the lambda branch, it fails. Am I doing something wrong?

ndavis-r7:
@jmanero-r7 any advice on this? Would like to close it out if possible

PR #65, opened 2015-11-11
Adds IAM Managed Policy

athompson-r7:
This takes advantage of the existing Policy mixin.

A Managed policy is very similar to a role policy in that there is
deny/allow statements.  The difference is that a Managed policy has one
and only one policy.

For example this is what a policy might look like:

    iam_managed_policy 'MyPolicy' do
      path '/my/path'
      allow do
        resource ARN
        action 's3:GetObject'
      end
      group 's3-users'
    end


jmanero-r7:
Looks good :+1:

PR #68, opened 2015-12-15
Closes #67; adds native support for ElastiCache ReplicationGroup reso…

### LINKED ISSUE #67: Support Elasticache Replication Group resource type
Convection supports Cache Clusters but not the Replication Group.

simonirwin-r7:
…urce type


athompson-r7:
Can you double check the types?  Some of these should be arrays I believe.

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticache-replicationgroup.html

athompson-r7:
@simonirwin-r7 Some of the properties are Arrays; we usually use accumulators here.  

Take a look at: `CacheSecurityGroupNames`, `PreferredCacheClusterAZs`, `SecurityGroupIds`, and `SnapshotArns`

PR #75, opened 2016-03-02
Change egress_rule signature

bturner-r7:
This changes the `egress_rule` method signature to take a destination.
You have always been able to set the destination of an egress rule using a block like:
```ruby
egress_rule(:tcp, 22) { self.destination = '0.0.0.0/0' }
```

This makes it a little easier by allowing you to set it in the method itself like:
```ruby
egress_rule(:tcp, 22, '0.0.0.0/0')
```

Additionally, this PR changes security rules to only specify a port if a port is declared.  This allows specifying rules that have no to/from port, like the following rule that is [recommended](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html) by Amazon for removing the default egress rule.
```ruby
egress_rule(-1, nil, '127.0.0.1/32')
```

### DIFF
--- a/example/security-groups.rb
+++ b/example/security-groups.rb
@@ -35,6 +35,43 @@ module Convection
 
         with_output
       end
+
+      ec2_security_group 'FoobarEgress' do
+        vpc stack.get('vpc', 'id')
+        description 'Foobar Ingress'
+
+        egress_rule(:tcp, 80, '0.0.0.0/0')
+        egress_rule(:tcp, 443, '0.0.0.0/0')
+
+        tag 'Name', "sg-foobar-egress-#{ stack.cloud }"
+        tag 'Service', 'foobar'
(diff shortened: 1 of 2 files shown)


fmitchell-r7:
:+1: Yay! Thanks for making egress rules easier to use.

bturner-r7:
I'm going to say I think rubocop is completely wrong on the "trailing comma" debate.  Ruby allows trailing commas in hashes, and it makes diff's cleaner when adding new items to a hash.  But I'll abide and remove the trailing comma: 63d9b0d1e73cd8dcf787d6bd2ec02a7ed430e765

simonirwin-r7:
Trailing commas are what made Puppet great

athompson-r7:
LGTM

PR #77, opened 2016-03-07
Add instance volumes

bturner-r7:
This adds support for adding volumes to EC2 instances.  It allows adding ephemeral instances, unencrypted EBS instances, and encrypted EBS instances.

The easiest way to see these changes is in the `example/foobar.rb` file.  I'll copy/paste the relevant/new sections here:

```ruby
# Create a KMS encryption key to encrypt the volume
kms_key 'FoobarKmsKey' do
  description 'Used to encrypt volumes'
  
  # don't delete the key when this stack is deleted
  deletion_policy 'Retain'

  policy do
    allow do
      sid 'Enable IAM User Permissions'
      principal :AWS => ["arn:aws:iam::#{MY_AWS_ACCOUNT_NUMBER}:root"]
      action 'kms:*'
      resource '*'
    end
  end
end

ec2_volume 'FoobarEncryptedVol' do
  availability_zone 'us-east-1a'
  size 20
  volume_type :gp2

  # encrypt with the key from this stack
  encrypted true
  kms_key fn_ref('FoobarKmsKey')

  # take a snapshot of the volume when this stack is deleted
  deletion_policy 'Snapshot'
end

ec2_instance 'FoobarWithEncryptedVol' do
  image_id stack['foobar-image']
  instance_type 'm3.medium'
  availability_zone 'us-east-1a'

  # mount the encrypted volume at /dev/xvdf
  volume do
    device '/dev/sdf'
    volume_id fn_ref('FoobarEncryptedVol')
  end

  # mount an ephemeral drive at /dev/xvdc
  block_device do
    device '/dev/sdc'
    virtual_name 'ephemeral0'
  end
end
```

The best way to review this is commit by commit:
* b21ac66009d87fb35abc9e5de09815a26f20eaf5 - adds support for the `DeletionPolicy` attribute.  This is supported by almost all resources and controls what happens to the resource when the stack is deleted.  This is helpful to allow encryption keys or important volumes to stick around after a stack is deleted.  See: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html
* 047598cac1f3cb6de8f4b136310f0a33126af873 - `kms_key` allows creating a new KMS encryption key and corresponding policy.  See: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-kms-key.html
* 7d6c93d019f15b86c3c43eb980ff8aaeb01549fd - `ec2_volume` allows creating a standalone EBS volume.  The volume can be attached to existing instances or new instances with the new `volume` attribute.  See: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-ebs-volume.html and http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html#cfn-ec2-instance-volumes
* 9b99b08b58e00834ad8532872d9f9bf7a8c4d73b - A new `block_device` allows attaching ephemeral volumes or net-new EBS volumes.  The main difference between block_device and volume is block_device creates the volume as part of launch, and volume attaches an existing volume.  See: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html#cfn-ec2-instance-blockdevicemappings

### DIFF
--- a/lib/convection/model/template/resource.rb
+++ b/lib/convection/model/template/resource.rb
@@ -354,7 +354,7 @@ module Convection
             'Properties' => properties.map(true, &:render)
           }.tap do |resource|
             resource['DependsOn'] = @depends_on unless @depends_on.empty?
-            resource['DeletionPolicy'] = @deletion_policy if @deletion_policy
+            resource['DeletionPolicy'] = @deletion_policy unless @deletion_policy.nil?
             render_condition(resource)
           end
         end
--- a/lib/convection/model/template/resource.rb
+++ b/lib/convection/model/template/resource.rb
@@ -295,6 +295,7 @@ module Convection
           @template = parent.template
           @type = self.class.type
           @depends_on = []
+          @deletion_policy = nil
           @exist = false
 
           ## Instantiate properties
@@ -316,6 +317,14 @@ module Convection
         def depends_on(resource)
           @depends_on << (resource.is_a?(Resource) ? resource.name : resource)
         end
+
+        # rubocop:disable Style/TrivialAccessors
+        #   We don't want to use an accessor (e.g. deletion_policy=) because
+        #   this is a DSL method
+        def deletion_policy(deletion_policy)
+          @deletion_policy = deletion_policy
+        end
+        # rubocop:enable Style/TrivialAccessors
 
         def reference
           {
@@ -345,6 +354,7 @@ module Convection
             'Properties' => properties.map(true, &:render)
           }.tap do |resource|
             resource['DependsOn'] = @depends_on unless @depends_on.empty?
+            resource['DeletionPolicy'] = @deletion_policy if @deletion_policy
             render_condition(resource)
           end
         end
--- a/lib/convection/model/template/resource_property/aws_ec2_block_store_block_device.rb
+++ b/lib/convection/model/template/resource_property/aws_ec2_block_store_block_device.rb
@@ -1,0 +1,22 @@
+require_relative '../resource_property'
+
+module Convection
+  module Model
+    class Template
+      class ResourceProperty
+        # Represents an {http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-template.html
+        # EC2 Block Store Block Device Property Type}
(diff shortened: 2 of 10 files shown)


eciramella-r7:
Can we rename the file from foobar.rb to something that reflects what's inside?

bturner-r7:
Sure, but that file was already there and named foobar, I just appended.  I'll rename it to something else.

lib/convection/model/template/resource.rb
```diff
@@ -345,6 +354,7 @@ def render
             'Properties' => properties.map(true, &:render)
           }.tap do |resource|
             resource['DependsOn'] = @depends_on unless @depends_on.empty?
+            resource['DeletionPolicy'] = @deletion_policy if @deletion_policy
```
fmitchell-r7: Any reason to not have this check for `nil` e.g.`unless @deletion_policy.nil?`

bturner-r7: No reason.  I switched to `nil` check.

bturner-r7:
I've renamed the `foobar` example stack to `instances` to more accurately reflect what's inside.

rvanburen-r7:
LGTM, but you might want to get more :eyes: on this.

lib/convection/model/template/resource_property/aws_ec2_block_store_block_device.rb
```diff
@@ -0,0 +1,22 @@
...
+        # EC2 Block Store Block Device Property Type}
+        class EC2BlockStoreBlockDevice < ResourceProperty
+          property :delete_on_termination, 'DeleteOnTermination'
+          property :encrypted, 'Encrypted'
+          property :iops, 'Iops'
+          property :snapshot, 'SnapshotId'
+          property :volume_size, 'VolumeSize'
+          alias size volume_size
+          property :volume_type, 'VolumeType'
+          alias type volume_type
```
fmitchell-r7: Are the `alias size` and `property :volume_type` lines supposed to be duplicated here?

fmitchell-r7:
:+1: Thanks for adding this!

PR #78, opened 2016-03-13
Add support for NAT gateways

bturner-r7:
This allows you to create a NAT gateways using AWS' new [NAT Gateway](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-nat-gateway.html) support.

A NAT Gateway looks like:
```ruby
# Create an Elastic IP that will be used by the NAT gateway
ec2_eip "NatGatewayIP1" do
  domain 'vpc'
end

# Create the NAT gateway.  The gateway must go in a public subnet
# that has an internet gateway attached.
ec2_nat_gateway "NatGateway1" do
  subnet fn_ref("TargetVPCSubnetPublic1")
  allocation_id get_att("NatGatewayIP1", 'AllocationId')
end

# Set the default route in the private route table to use the NAT Gateway
ec2_route "NatGatewayRoute1" do
  destination '0.0.0.0/0'
  nat_gateway fn_ref("NatGateway1")
  route_table_id private_table
end
```

### DIFF
--- a/example/vpc.rb
+++ b/example/vpc.rb
@@ -15,6 +15,10 @@ module Convection
         tag 'Stack', stack.cloud
         with_output 'id'
 
+        #
+        # PUBLIC SUBNETS
+        #
+
         ## Add an InternetGateway
         add_internet_gateway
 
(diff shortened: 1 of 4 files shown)


simonirwin-r7:
LGTM

Do you think this should be our default behaviour?  Any reason we might want to support EC2 instance NAT gateways in future?  I might make move vpc_builder to use NAT gateways by default.

dgreene-r7:
:+1:

Thanks for adding the EIP resource too!

PR #79, opened 2016-04-07
tag support for elasticache cluster and replication group

bcall-r7:
This pr is for adding the mixin and render method to add tag support to elasticache clusters and elasticache replication groups


athompson-r7:
Looks good. Thanks @bcall-r7!

PR #81, opened 2016-04-11
Asg update policy

simonirwin-r7:


simonirwin-r7:
Based on what I've been doing to better support Lambda, I think I can rework this support for UpdatePolicy.

athompson-r7:
Seems reasonable.  :shipit: 

Can you just make sure any enhancements you are thinking of make it into issues.  Thanks @simonirwin-r7

PR #84, opened 2016-04-12
Fix support for Lambdas

simonirwin-r7:
Added full support for S3 bucket reference, and VPC integration.  Unit test added, showing we can support:

```
      lambda_function 'TestLambda' do
        description 'Test description'
        handler 'index.handler'
        runtime 'nodejs'
        role 'arn:aws:x:y:z'

        code do
          s3_bucket 'testbucket'
          s3_key 'testkey'
        end

        vpc_config do
          security_groups %w(group1 group2)
          subnets %w(subnet1a subnet1b)
        end
      end
```


athompson-r7:
Thanks for fixing this @simonirwin-r7

PR #88, opened 2016-04-14
Support VpcEndpoint

simonirwin-r7:
This is the proposed API:

```
ec2_vpc_endpoint 'TestVpcEndpoint' do
  service 's3'
  vpc 'vpc-foo'
  route_tables %w(table1 table2)

  allow do
    s3_resource 'bucket-bar', '*'
    action 's3:GetObject'
  end
end
```

### DIFF
--- a/test/convection/model/test_vpc_endpoint.rb
+++ b/test/convection/model/test_vpc_endpoint.rb
@@ -27,7 +27,6 @@ class TestVpcEndpoint < Minitest::Test
   def test_endpoint
     json = from_json['Resources']
 
-    pp json
     endpoint = json['TestVpcEndpoint']
     refute endpoint.nil?, 'VpcEndpoint not present in generated template'
 
@@ -42,6 +41,9 @@ class TestVpcEndpoint < Minitest::Test
     refute s3_string_arr.nil?, 'ServiceName value is not specified as a string array'
 
     assert s3_string_arr.is_a? Array
-    assert s3_string_arr[1][2] == 's3', 'ServiceName does not contain \'s3\' as final element'
+    s3_path_as_array = s3_string_arr[1]
+    refute s3_path_as_array.nil?, "The path for S3 should be defined as an array in #{s3_string_arr}"
+
+    assert s3_path_as_array.last == 's3', "ServiceName (#{s3_path_as_array}) does not contain \'s3\' as final element"
   end
 end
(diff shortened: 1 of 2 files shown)


alee-r7:
How will this work in razor-convection to pull out the route table id from a particular subnet on a vpc,  to feed it in the api?

test/convection/model/test_vpc_endpoint.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def from_json
+    JSON.parse(@template.to_json)
+  end
+
+  def test_endpoint
+    json = from_json['Resources']
+
+    pp json
```
erran-r7: :smirk:

test/convection/model/test_vpc_endpoint.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    service_name = props['ServiceName']
+    refute service_name.nil?, 'ServiceName not present in generated template'
+
+    assert service_name.is_a? Hash
+    s3_string_arr = service_name['Fn::Join']
+    refute s3_string_arr.nil?, 'ServiceName value is not specified as a string array'
+
+    assert s3_string_arr.is_a? Array
+    assert s3_string_arr[1][2] == 's3', 'ServiceName does not contain \'s3\' as final element'
```
erran-r7: Consider creating a variable for the value of `s3_string_arr[1]` which would be useful in debugging failures?

simonirwin-r7:
@alee-r7 the `vpc` template would need to integrate this change.  I'm not sure if we want all, or a subset of, subnets to use a VPC Endpoint.  Furthermore, does it mean we have to add a policy for EVERY bucket to the VPC Endpoint?  I haven't fully read the docs, admittedly.

I'm not even sure if the razor-prod-1 vpc stack is still managed by convection.

alee-r7:
@simonirwin-r7: Yeah so 2 options we have. Create a universal vpc endpoint within the vpc for all subnets to use the endpoint. This will probably mean the endpoint policy will allow everything.

An alternative is to create a specific vpc endpoint for just one or two subnets that have a more restrictive policy. This approach would can be more focused on the particular feature and thus can be defined in a non vpc stack like assetui for example. Of course, this would mean assetui would have to ask the vpc stack for the route table id for subnet veyron.

alee-r7:
:ship: 

I think we should give it a try.

PR #91, opened 2016-04-19
Create an AWS::Events::Rule resource

erran-r7:
Adds a resource for `AWS::Events::Rule` to enable Convection support of creating CloudWatch event rules.

See also [AWS::Events::Rule](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-rule.html#cfn-events-rule-eventpattern).


erran-r7:
Adding the **do not merge** label since we should create a Target helper class/method for adding targets.

erran-r7:
Removing do not merge since I've created a resource property for event rule targets in bc60620. One thing to not is that we don't have resource properties for [events/event patterns](http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/CloudWatchEventsandEventPatterns.html) but I believe this is valid since we won't need to create events as of yet and an event pattern is created as a Hash of arrays. The format for specifying that pattern today would be:

```ruby
event_pattern 'detail-type', ['EC2 Instance State-change Notification']
event_pattern 'detail', { 'state' => ['running'] }
```

Each key/value pair would be appended to the pattern array which matches the cloud formation schema. Perhaps in the future we can make this prettier. Any suggestions?

erran-r7:
# Example
```ruby
cw_event = ::Convection.template do
  events_rule 'ScheduledRule' do
    description 'A scheduled rule.'
    domain 'ScheduledRuleDomain'
    name 'ScheduledRuleName'
    role_arn 'role_arn'
    schedule_expression 'schedule_expression'
    state 'state'

    event_pattern 'detail-type', ['EC2 Instance State-change Notification']

    target do
      arn 'deadb33f'
      id 'deadb33f'
    end
  end
end
```

erran-r7:
Oops, rebased `master` instead of `v0.2`...

erran-r7:
Please see https://github.com/rapid7/convection/pull/91 instead.

PR #92, opened 2016-04-19
Generic trust relationship

simonirwin-r7:
Within an `iam_role` definition, I can now specify:

`trust_ec2_instances`
or
`trust_service 'ec2'`

This means any other services can be supported without an explicit `trust_XXX` method. e.g.

`trust_service 'lambda'`

I need the Lambda trust method above.  This will generate:

```
    "AssumeRolePolicyDocument": {
          "Version": "2012-10-17",
          "Statement": [
            {
              "Effect": "Allow",
              "Action": [
                "sts:AssumeRole"
              ],
              "Principal": {
                "Service": "lambda.amazonaws.com"
              }
            }
          ]
        }
```

### DIFF
--- a/test/convection/model/test_trust.rb
+++ b/test/convection/model/test_trust.rb
@@ -1,0 +1,29 @@
+require 'test_helper'
+require 'json'
+require 'pp'
+
+class TestTrust < Minitest::Test
+  def setup
+    @template = ::Convection.template do
+      description 'Trust Test Template'
+
+      iam_role 'FooRole' do
+        trust_service 'bar'
+      end
+    end
+  end
+
+  def from_json
+    JSON.parse(@template.to_json)
+  end
+
+  def test_trust
+    json = from_json['Resources']['FooRole']['Properties']
+    doc = json['AssumeRolePolicyDocument']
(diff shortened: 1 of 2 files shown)


athompson-r7:
This is great @simonirwin-r7!  Can you add some tests that show this off?

jmanero-r7:
Cool!

erran-r7:
@simonirwin-r7 @jmanero-r7 @athompson-r7 thoughts on converting the ec2 method to be implemented using the new generic method?

athompson-r7:
@simonirwin-r7 Looks good :shipit: 
@erran-r7 I'm for it, let's split that work out into a separate ticket.

erran-r7:
> @erran-r7 I'm for it, let's split that work out into a separate ticket.

👍 I'm happy with that.

test/convection/model/test_trust.rb
```diff
@@ -0,0 +1,29 @@
+require 'test_helper'
+require 'json'
+require 'pp'
```
erran-r7: 😏

PR #94, opened 2016-04-20
Reimplement AWS trust service methods after merging #92

### LINKED ISSUE #93: Use generic trust service method in the trust_ec2_resources method
A generic trust method was introduced in #92.

erran-r7:
Closes #93.

---

~~**NOTE**: These changes will update the names for most of these policies.~~

*edit*: Changes to maintain the policy names were implemented.

### DIFF
--- a/lib/convection/model/template/resource/aws_iam_role.rb
+++ b/lib/convection/model/template/resource/aws_iam_role.rb
@@ -48,7 +48,6 @@ module Convection
           ## Add a canned trust policy for EMR
           def trust_emr(&block)
             trust_service('elasticmapreduce', 'trust-emr', &block)
-            trust_service('', '', &block)
           end
 
           ## Add a canned trust policy for Cloudtrail
--- a/lib/convection/model/template/resource/aws_iam_role.rb
+++ b/lib/convection/model/template/resource/aws_iam_role.rb
@@ -25,8 +25,9 @@ module Convection
           end
 
           ## Add a canned trust policy for any AWS service
-          def trust_service(name, &block)
-            @trust_relationship = Model::Mixin::Policy.new(:name => "trust-#{name}-service", :template => @template)
+          def trust_service(name, policy_name = nil, &block)
+            policy_name ||= "trust-#{name}-service"
+            @trust_relationship = Model::Mixin::Policy.new(:name => policy_name, :template => @template)
             trust_relationship.allow do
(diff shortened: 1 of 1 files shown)


simonirwin-r7:
@erran-r7 I don't *think* the policy name (which will change as part of this PR) actually gets used.

Here's a snippet of a CF template generated by Convection:

```
"AssumeRolePolicyDocument": {
          "Version": "2012-10-17",
          "Statement": [
            {
              "Effect": "Allow",
              "Action": [
                "sts:AssumeRole"
              ],
              "Principal": {
                "Service": "ec2.amazonaws.com"
              }
            }
          ]
        }
```

athompson-r7:
I don't like the idea of the name changing.  Anything we can do to avoid this?

erran-r7:
@athompson-r7 Yes, simply add a name parameter to the generic message that isn't required. It'll default to the `"trust-#{service}-service"` name format (new) but we can pass in the old names to ensure consistency when using these methods.

erran-r7:
> I don't like the idea of the name changing. Anything we can do to avoid this?



> @athompson-r7 Yes, simply add a name parameter to the generic message that isn't required. It'll default to the "trust-#{service}-service" name format (new) but we can pass in the old names to ensure consistency when using these methods.

@athompson-r7 check out 3932f60 where we keep parity with the existing methods by maintaining the policy names.

lib/convection/model/template/resource/aws_iam_role.rb
```diff
@@ -36,49 +37,23 @@ def trust_service(name, &block)
...
           def trust_emr(&block)
-            @trust_relationship = Model::Mixin::Policy.new(:name => 'trust-emr', :template => @template)
-            trust_relationship.allow do
-              action 'sts:AssumeRole'
-              principal :Service => 'elasticmapreduce.amazonaws.com'
-            end
-            trust_relationship.instance_exec(&block) if block
-            trust_relationship
+            trust_service('elasticmapreduce', 'trust-emr', &block)
+            trust_service('', '', &block)
```
erran-r7: eek! 🍝

erran-r7: Fixed in https://github.com/rapid7/convection/pull/94/commits/d857719fbeff7a58ed16b95c2616f9d3145f937a.

PR #98, opened 2016-04-26
VPC Endpoint error when generating the S3 service name

simonirwin-r7:
The join was creating something like this:

`com.amazonaws.us-east-1s3`

It should be:

`com.amazonaws.us-east-1.s3`


athompson-r7:
LGTM

PR #117, opened 2016-05-03
Resolve linting failures in DSL helpers and resource templates

erran-r7:


erran-r7:
Looks like `bundle exec rake` does not run with the same profile that travisci does weirdly.

PR #119, opened 2016-05-04
Support stack life cycle tasks (see also #114)

### LINKED ISSUE #114: Support before/after stack creation and deletion tasks
Tasks can utilise information around the stack object and be used to perform operations in AWS or elsewhere as the user sees fit.

Today at Rapid7 we have internal support for this via monkey patching the stack create/delete methods to enumerate a registry of tasks to execute.

# Subtasks
- [x] Create a `before_create` task type.
- [x] Create a `after_create` task type.
- [x] Create a `before_delete` task type.
- [x] Create a `after_delete` task type.

erran-r7:
Fix #114.


erran-r7:
**NOTE**: This feature is a net new feature so would not be breaking compatibility but we should define whether it should be added in the next major version based on it being a large change.

lib/convection/control/stack.rb
```diff
@@ -359,6 +402,14 @@ def cf_tags
           }
         end
       end
+
+      def hash_with_array_as_default
+        Hash.new { |hash, key| hash[key] = [] }
+      end
+
+      def reversed_hash(hash)
```
erran-r7: `Stack#reversed_hash` and `Stack#hash_with_array_as_default` are no longer used.

test/convection/tasks/test_ec2_vpc.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def test_before_create_tasks
+    cf_client = Minitest::Mock.new
+    ec2_client = Minitest::Mock.new
+    Aws::CloudFormation::Client.stub :new, cf_client do
+      any_args = [->(*) { true }]
+      cf_client.expect(:create_stack, nil, any_args)
+      cf_client.expect(:describe_stacks, nil)
+      def cf_client.describe_stacks(*)
```
erran-r7: These could go in a `mock_cf_client` helper method.

test/convection/tasks/test_ec2_vpc.rb
```diff
@@ -0,0 +1,59 @@
...
+        raise Aws::CloudFormation::Errors::ValidationError.new('foo', 'bar')
+      end
+
+      Aws::EC2::Client.stub :new, ec2_client do
+        task = CollectAvailabilityZonesTask.new
+        stack = ::Convection::Control::Stack.new('EC2 VPC Test Stack', @template) do
+          before_create_task task
+        end
+        def stack.availability_zones
+          %w(foo bar baz)
```
erran-r7: Be reasonable, these obviously aren't legit zones. 😉

test/convection/tasks/test_ec2_vpc.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def test_before_create_tasks
+    cf_client = Minitest::Mock.new
+    ec2_client = Minitest::Mock.new
+    Aws::CloudFormation::Client.stub :new, cf_client do
+      any_args = [->(*) { true }]
+      cf_client.expect(:create_stack, nil, any_args)
+      cf_client.expect(:describe_stacks, nil)
+      def cf_client.describe_stacks(*)
+        raise Aws::CloudFormation::Errors::ValidationError.new('foo', 'bar')
```
erran-r7: foo, bar, really?

lib/convection/control/stack.rb
erran-r7: Resolved in 1793dff3f2587ed7d38766e1c7667f977de123cf.

lib/convection/control/stack.rb
```diff
@@ -248,6 +275,22 @@ def validate
         puts "\nTemplate validated successfully"
       end
 
+      def after_create_task(task)
+        @tasks[:after_create].unshift(task)
```
simonirwin-r7: if i define 2 tasks:

```
stack test, MY::TEST::STACK do
  after_create_task(a)
  after_create_task(b)
end
```

then the order of task execution would be `b` then `a`?

erran-r7: That's correct. 👍

simonirwin-r7: isn't that counterintuitive?

erran-r7: For after tasks, I disagree. Before tasks are run in the order you register them. I've modelled tasks in a manner similar to how test hooks are dealt with in several unit testing frameworks. If you have cleanup you do it in reverse order, from setup, in my experience.

erran-r7: @simonirwin-r7 @abegley-r7 and myself discussed this during our daily scrum. We're going to run after tasks in the order they are registered and document that users should define accordingly them if they have a execution dependent order.

erran-r7: @simonirwin-r7 @abegley-r7 @erran-r7 resolved in 84bc2cf04381c5042d0716634b677a1b9a2f6f62.

erran-r7:
@simonirwin-r7 @abegley-r7 see https://github.com/rapid7/convection/pull/120 instead for these changes. I mistakenly merged master into this branch.

PR #120, opened 2016-05-05
Support stack life cycle tasks (#114 and #119)

erran-r7:
See also #114 (original issue) and #119 (accidentally merged/pushed a merge from an incorrect base branch).

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -23,7 +23,7 @@ Metrics/AbcSize:
 # Offense count: 3
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 267
+  Max: 298
 
 # Offense count: 6
 Metrics/CyclomaticComplexity:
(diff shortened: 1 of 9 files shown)


erran-r7:
As I'm doing some testing internally I've noticed: the lack of output for tasks is pretty confusing.

simonirwin-r7:
Think this looks good.

erran-r7:
Looks like a new version of RuboCop pulled in by TravisCI started causing failures here. See https://github.com/rapid7/convection/pull/123 for the resolved violations (nothing in this PR is in violation currently).

erran-r7:
@athompson-r7 @jmanero-r7 Any additional feedback here? Does it make sense to merge this feature into the `v0.2` branch? Should we create a new `v0.3` branch/version for it since it is a pretty large change? I'm not sure when it'd make sense to start creating a `v1.0.0.pre.x` version to start iterating on.

PR #123, opened 2016-05-10
Resolve violations introduced by pulling in RuboCop v0.40.0

erran-r7:
@athompson-r7 @fmitchell-r7 @jmanero-r7 @simonirwin-r7 @eciramella-r7 Looks like our version constraint on RuboCop is too light since this is the second time I've seen RuboCop release a new version and existing files (which passed linting) start failing the TravisCI build.

This resolves the 15 violations that started failing one of my PRs.

PR #125, opened 2016-05-10
👮🏽 Constrain RuboCop to ~> 0.40.0

### LINKED ISSUE #124: Be more explicit with the Rubocop version
We should lock our Rubocop version at the patch level to prevent minor version bumps.  This is in reference to #123.

erran-r7:
This will prevent TravisCI builds from failing when RuboCop rules change
drastically in MINOR versions. PATCH versions will still be allowed to
be pulled automatically.

👮 👮 👮 

Fix #124.
See also #123.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ group :development do
   gem 'bundler', '~> 1.7'
   gem 'minitest'
   gem 'rake', '~> 10.0'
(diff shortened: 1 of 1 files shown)


abegley-r7:
LGTM 👍

athompson-r7:
LGTM

Issue #110, opened 2016-04-27
Improve performance of performing diffs


erran-r7:
See also #109 which is a research spike where we'll document any reasons diffs may currently be slow.

athompson-r7:
#128 May be a solution to this, depending what you are doing.

Also see my notes about splay in #111

erran-r7:
@bcall-r7 were this (and #109) solved by #187?

erran-r7:
Closing due to issue age.

PR #128, opened 2016-05-11
Feature: Diff individual stacks

fmitchell-r7:
This makes the `convection diff` command behave the same way the `convection converge` command does. It takes an optional `STACK` argument, the name of the stack to diff up to. If the stack name's given, the diff will halt at the named stack.

### DIFF
--- a/bin/convection
+++ b/bin/convection
@@ -26,15 +26,15 @@ module Convection
       end
     end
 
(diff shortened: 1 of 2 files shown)


jmanero-r7:
👍 Looks good

erran-r7:
:shipit: looks good to me. As @athompson-r7 pointed out this will help out with #110. cc @bturner-r7

PR #131, opened 2016-06-02
Fix #130 - Support after_update/before_update task types

### LINKED ISSUE #130: Support before/after update tasks around stack updates
- [ ] Create a new `before_update` task category
- [ ] Create a new `after_update` task category
- [ ] Run update tasks before/after the update phase

See also https://github.com/rapid7/convection/issues/114 and https://github.com/rapid7/convection/issues/120.

erran-r7:
Fix #130:

>- [x] Create a new `before_update` task category
>- [x] Create a new `after_update` task category
>- [x] Run update tasks before/after the update phase
>
>See also https://github.com/rapid7/convection/issues/114 and https://github.com/rapid7/convection/issues/120.

### DIFF
--- a/lib/convection/control/stack.rb
+++ b/lib/convection/control/stack.rb
@@ -84,7 +84,7 @@ module Convection
         @id = nil
         @outputs = {}
         @resources = {}
-        @tasks = { after_create: [], after_delete: [], before_create: [], before_delete: [] }
+        @tasks = { after_create: [], after_delete: [], after_update: [], before_create: [], before_delete: [], before_update: [] }
         instance_exec(&block) if block
         @current_template = {}
         @last_event_seen = nil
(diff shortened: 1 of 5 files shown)


erran-r7:
I fixed the 👮 violation in 9a88cc9. Local rake passes. Clearly wasn't running the whole rake suite (just tests) earlier.

lib/convection/control/stack.rb
```diff
@@ -178,7 +178,16 @@ def apply(&block)
           o[:capabilities] = capabilities
         end
 
-        if exist?
+        ## Execute before create tasks
+        before_task_type = exist? ? :before_update : :before_create
+        @tasks[before_task_type].delete_if do |task|
```
simonirwin-r7: I think that we should only call the `before_update` task *if* the stack has changes to be applied

erran-r7: @simonirwin-r7 good point! I agree and overlooked it initially. 👍

erran-r7: Resolved in c5b96224168ac559bba02c48d4cdde29b5b596bb.

PR #132, opened 2016-06-06
Fix #129 - Add an assume_role_policy DSL method to iam_role

### LINKED ISSUE #129: Make custom trust relationships easier
This should be simpler than writing out the following:

```ruby
iam_role 'CloudHealthIAMRole' do
  # You must define a policy specifically and use "self" from within the `iam_role` DSL
  # method to define a different policy for the trust relationship.
  self.trust_relationship = Convection::Model::Mixin::Policy.new(:name => "CloudHealth-assumerole", :template => @template)

  trust_relationship.allow do
    action 'sts:AssumeRole'
    principal :AWS => "arn:aws:iam::#{CLOUDHEALTH_ACCOUNT}:root"
    condition :StringEquals => { "sts:ExternalId" => CLOUDHEALTH_EXTERNALID }
  end

  ...
end
```

erran-r7:
Fix #129

### DIFF
--- a/lib/convection/model/template/resource/aws_iam_role.rb
+++ b/lib/convection/model/template/resource/aws_iam_role.rb
@@ -6,6 +6,11 @@ module Convection
       module Resource
         ## Role DSL
         module IAMRole
+          def assume_role_policy(policy_name, &block)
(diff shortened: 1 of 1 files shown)


fmitchell-r7:
👍 Thanks for making this easier! :shipit:

PR #133, opened 2016-06-10
Support vpn

simonirwin-r7:
I'd like to revisit this at some stage to add some sugar to more easily link up the different resources.

Need to add an example to the docs, too.

### DIFF
--- a/lib/convection/model/template/resource/aws_ec2_customer_gateway.rb
+++ b/lib/convection/model/template/resource/aws_ec2_customer_gateway.rb
@@ -11,7 +11,7 @@ module Convection
           include Model::Mixin::Taggable
 
           type 'AWS::EC2::CustomerGateway'
-          property :conn_type, 'Type'
+          property :connection_type, 'Type'
           property :bgp_asn, 'BgpAsn'
           property :ip_address, 'IpAddress'
 
--- a/lib/convection/model/template/resource/aws_ec2_customer_gateway.rb
+++ b/lib/convection/model/template/resource/aws_ec2_customer_gateway.rb
@@ -1,0 +1,27 @@
+require_relative '../resource'
+
+module Convection
+  module Model
+    class Template
+      class Resource
+        ##
+        # AWS::EC2::CustomerGateway
+        ##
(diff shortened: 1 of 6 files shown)


lib/convection/model/template/resource/aws_ec2_vpn_connection.rb
```diff
@@ -0,0 +1,28 @@
...
+    class Template
+      class Resource
+        ##
+        # AWS::EC2::VPNConnection
+        ##
+        class EC2VPNConnection < Resource
+          include Model::Mixin::Taggable
+
+          type 'AWS::EC2::VPNConnection'
+          property :conn_type, 'Type'
```
erran-r7: Why use `:conn_type`? Let's use `connection_type` since `type` is taken. It's more clear.

lib/convection/model/template/resource/aws_ec2_customer_gateway.rb
```diff
@@ -0,0 +1,27 @@
...
+    class Template
+      class Resource
+        ##
+        # AWS::EC2::CustomerGateway
+        ##
+        class EC2CustomerGateway < Resource
+          include Model::Mixin::Taggable
+
+          type 'AWS::EC2::CustomerGateway'
+          property :conn_type, 'Type'
```
erran-r7: > Why use `:conn_type`? Let's use `connection_type` since `type` is taken. It's more clear.

lib/convection/model/template/resource/aws_ec2_vpn_gateway.rb
```diff
@@ -0,0 +1,25 @@
...
+    class Template
+      class Resource
+        ##
+        # AWS::EC2::VPNGateway
+        ##
+        class EC2VPNGateway < Resource
+          include Model::Mixin::Taggable
+
+          type 'AWS::EC2::VPNGateway'
+          property :conn_type, 'Type'
```
erran-r7: > Why use `:conn_type`? Let's use `connection_type` since `type` is taken. It's more clear.

PR #135, opened 2016-06-27
Fix a typo in the aws_s3_bucket.rb resource

erran-r7:
See also https://github.com/rapid7/convection/commit/782208968f#diff-7126ecd8a0f873d6bdeffcaf269fb9a7R15 where this typo was introduced.

cc @jmanero-r7 @athompson-r7 @simonirwin-r7 @abegley-r7


erran-r7:
@jmanero-r7 @bturner-r7 @abegley-r7 as discussed adhoc — d24cfaf resolves the issue of any previous uses of the method.

PR #136, opened 2016-06-27
Support s3 bucket configuration types

erran-r7:
These changes are required to support setting the CorsConfiguration through convection.

---

cc @abegley-r7 @simonirwin-r7 @mmckeaveney-r7

### DIFF
--- a/lib/convection/model/template/resource/aws_s3_bucket.rb
+++ b/lib/convection/model/template/resource/aws_s3_bucket.rb
@@ -19,6 +19,12 @@ module Convection
           property :notification_configuration, 'NotificationConfiguration'
           property :versioning_configuration, 'VersioningConfiguration'
 
+          def cors_configuration(&block)
+            config = ResourceProperty::S3CorsConfiguration.new(self)
+            config.instance_exec(&block) if block
+            properties['CorsConfiguration'].set(config)
+          end
+
           def cors_configurationm(*args)
             warn 'DEPRECATED: "cors_configurationm" is deprecated. Please use "cors_configuration" instead. https://github.com/rapid7/convection/pull/135'
             cors_configuration(*args)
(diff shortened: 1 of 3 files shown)


lib/convection/model/template/resource/aws_s3_bucket.rb
```diff
@@ -13,12 +13,18 @@ class S3Bucket < Resource
           type 'AWS::S3::Bucket'
           property :bucket_name, 'BucketName'
           property :access_control, 'AccessControl'
-          property :cors_configurationm, 'CorsConfiguration'
+          property :cors_configuration, 'CorsConfiguration'
```
erran-r7: **NOTE**: This fix should be pulled in via #135.

lib/convection/model/template/resource_property/aws_s3_cors_configuration_rule.rb
```diff
@@ -0,0 +1,20 @@
+require_relative '../resource_property'
+
+module Convection
+  module Model
+    class Template
+      class ResourceProperty
+        # Represents an {http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-cors-corsrule.html
+        # Amazon S3 Cors Configuration Rule}
+        class S3CorsConfiguration < ResourceProperty
```
erran-r7: 😏 small typo here.

abegley-r7:
Looks good, only concern is impacting current stacks, spoke with @e with regards to current usage, looks like no other stack references/makes use of CORS.

erran-r7:
> Looks good, only concern is impacting current stacks, spoke with @erran-r7 with regards to current usage, looks like no other stack references/makes use of CORS.

@abegley-r7 spoke to @jmanero-r7 and I'm going to address that in #135 with a deprecation warning for the bad property name.

erran-r7:
> I'm going to address that in #135 with a deprecation warning for the bad property name.

Done! Merging after tests complete.

PR #138, opened 2016-07-01
Bring v0.2 down into master

eciramella-r7:
To prevent confusion and speed adoption, bringing v0.2 down into master.


bcall-r7:
ship it

PR #139, opened 2016-07-02
Add information around versioning/branching

erran-r7:
Thoughts @jmanero-r7 @athompson-r7 @bcall-r7?


athompson-r7:
The language seems fine to me.

PR #153, opened 2016-07-12
Fix #152 - Add support for the Metadata CloudFormation template section

### LINKED ISSUE #152: Add support for Metadata to template
See also [Metadata](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/metadata-section-structure.html).

As I was working on #151 I realised we don't actually support the CloudFormation Metadata section for templates. We should add this and make sure we label it as a "resource dependent change".

erran-r7:
Fix #152.

# After
```ruby
require 'convection'

module Templates
  VPC = Convection.template do
    description 'A dummy VPC to demo.'

    # metadata key/value pairs can be defined using Template#metadata(KEY, VALUE)
    metadata 'Instances', {
      'Description' => 'Information about the instances'
    }

    ec2_vpc 'TargetVPC' do
      network '127.123.0.0/16'
      subnet_length 24
      enable_dns

      with_output
    end
  end
end
```


bcall-r7:
Ship it

PR #164, opened 2016-07-21
Adding new feature guide

bcall-r7:


erran-r7:
:shipit: On the ball, let's merge this after you push the unit test coverage you mentioned in a private message to me.

docs/adding-new-resource-coverage.md
```diff
@@ -0,0 +1,269 @@
...
+9. Add the mock template. Below is a example
+```ruby
+require 'spec_helper'
+
+class Convection::Model::Template::Resource
+  describe DHCPOptions do
+    let(:template) do
+      Convection.template do
+
+        ec2_dhcp_options 'TestOptions' do
```
erran-r7: This actually wouldn't work right now. Currently the `dsl_name` would default to `dhcp_options` since we didn't include `EC2` in the class name. We can either do that or pass in the second parameter `dsl_name` to the `type` call in the `DHCPOptions` class.

erran-r7: Whoa, I'm totally wrong. Looking at the `type` DSL method it does the following:

```ruby
@name = dsl_name || DSL::Helpers.method_name(cf_type)
```

Swore we passed `self.class` in vs. `cf_type`. We don't actually need to update the method name here. Can you double check for consistency between other classes though? The file name and class names definitely don't match.

PR #165, opened 2016-07-22
Add aws rds event subscription resource type support

erran-r7:


bcall-r7:
Could you add some Specs? Thanks

erran-r7:
@bcall-r7 there aren't any complex resource properties here so I'm going to go ahead and merge without adding specs (they'd only be adding line coverage not useful testing).

PR #168, opened 2016-07-26
Feature/converge select stacks and stackgroups issue 108 126

bcall-r7:


lib/convection/control/cloud.rb
```diff
@@ -20,45 +20,105 @@ def deck
...
-      def converge(to_stack, &block)
-        unless to_stack.nil? || stacks.include?(to_stack)
-          block.call(Model::Event.new(:error, "Stack #{ to_stack } is not defined", :error)) if block
-          return
-        end
+      def stack_groups
+        @cloudfile.stack_groups
+      end
 
+      def create_structure(list)
```
bcall-r7: im going to look into collecting only the stacks we want instead of iterating over all to find them

lib/convection/control/cloud.rb
```diff
@@ -1,6 +1,6 @@
 require_relative '../model/cloudfile'
 require_relative '../model/event'
-
+require 'pp'
```
erran-r7: :trollface:

lib/convection/control/cloud.rb
```diff
@@ -20,45 +20,105 @@ def deck
...
+        end
+      end
 
-          difference.each { |diff| block.call(diff) }
+      def apply_converge(stack, &block)
+        block.call(Model::Event.new(:converge, "Stack #{ stack.name }", :info)) if block
+        stack.apply(&block)
+        if stack.error?
+          block.call(Model::Event.new(:error, "Error converging stack #{ stack.name }", :error), stack.errors) if block
+          return false
```
erran-r7: @bcall-r7 pretty sure `stack.success?` below will return false anyways, in which case this isn't needed since we don't have any other logic below.

lib/convection/control/cloud.rb
```diff
@@ -20,45 +20,105 @@ def deck
...
-          if stack.error?
-            block.call(Model::Event.new(:error, "Error converging stack #{ stack.name }", :error), stack.errors) if block
-            break
-          end
+          hash[stack.name] = stack if list.include?(stack.name.to_s)
+        end
+        hash
+      end
 
+      def simple_converge(to_stack, &block)
```
erran-r7: How about `converge_until`?

lib/convection/control/cloud.rb
```diff
@@ -20,45 +20,105 @@ def deck
...
+        deck.each do |stack|
+          break unless apply_converge(stack, &block)
           ## Stop on converge error
           break unless stack.success?
-
           ## Stop here
           break if !to_stack.nil? && stack.name == to_stack
```
erran-r7: If this is the only difference, I wonder if we could make these more similar.

lib/convection/control/cloud.rb
bcall-r7: haters gonna hate

Issue #126, opened 2016-05-11
Support the notion of "stack groups" which can be used to logically identify multiple stacks


erran-r7:
This pairs nicely with #108.

erran-r7:
Fixed by #169.

Issue #108, opened 2016-04-27
Support converging select stacks


erran-r7:
Currently `#converge` supports covering **up to** a specific stack, but will converge **all stacks up to** that point.

This change will support converging only stacks specified (a list or single) on the command line.

erran-r7:
Fixed by #169.

PR #169, opened 2016-07-26
Support stack groups and listing specific stacks to converge

bcall-r7:
# References (see also)
* #108 — Support converging select stacks
* #126 — Support the notion of "stack groups" which can be used to logically identify multiple stacks
* #168 — Support stack groups and listing specific stacks to converge

# Verification Steps
- Stack groups
  - [x] Diff with a defined stack group works (converges only the stacks defined in the group)
  - [x] Converge with a defined stack group works (diffs only the stacks defined in the group)
  - [x] Diff with an undefined stack group works (emits an error and does not diff)
  - [ ] Converge with an undefined stack group works (emits an error and does not converge)
- Stack lists
  - [x] Diff with a list of defined stacks works
  - [x] Converge with a list of defined stacks works
  - [x] Diff with a valid stack and undefined stack in the list works (diffs and emits an error for the bad stack)
  - [ ] Converge with a valid stack and undefined stack in the list works (converges and emits an error for the bad stack)
  - [x] Diff with only an undefined stack in the list works (diffs and emits an error for the bad stack)
  - [ ] Converge with only an undefined stack in the list works (converges and emits an error for the bad stack)
    - **NOTE**: This converges valid stacks but does not emit errors for bad stack names.
- "Legacy" operations (up to a given stack)
  - [x]  Legacy diffs work (up to a given stack)
  - [x]  Legacy Converge works (up to a given stack)
  - [ ] Diff with stack that is undefined errors correctly (prints error and does not diff)
  - [ ] Converge with stack that is undefined errors correctly (prints error and does not converge)


bin/convection
```diff
@@ -48,6 +51,8 @@ module Convection
             last_event = nil
           end
           say_status(*d.to_thor)
+        elsif d.name == :error
```
erran-r7: Won't `d.is_a?(Model::Event)` above return true meaning we never enter this?

lib/convection/control/cloud.rb
```diff
@@ -16,17 +16,21 @@ def stacks
...
       end
 
-      def converge(to_stack, &block)
-        unless to_stack.nil? || stacks.include?(to_stack)
-          block.call(Model::Event.new(:error, "Stack #{ to_stack } is not defined", :error)) if block
-          return
-        end
+      def filter_deck(options = {}, &block)
+        included_stacks = stack_groups[options[:stack_group]]
+        included_stacks ||= options.fetch(:stack_list, [])
```
erran-r7: Since the default value isn't returned here in any case we can just use `Hash#[]` instead of trying to use fetch's default option.

lib/convection/control/cloud.rb
```diff
@@ -16,17 +16,21 @@ def stacks
...
-        unless to_stack.nil? || stacks.include?(to_stack)
-          block.call(Model::Event.new(:error, "Stack #{ to_stack } is not defined", :error)) if block
-          return
-        end
+      def filter_deck(options = {}, &block)
+        included_stacks = stack_groups[options[:stack_group]]
+        included_stacks ||= options[:stack_list]
+        return @cloudfile.deck if included_stacks.nil?
+        stack_list = @cloudfile.stacks.map { |name, stack| stack if included_stacks.include?(name) }.compact
+        block.call(Model::Event.new(:error, "Stack #{included_stacks} not exist", :error))
```
erran-r7: Call this regardless? 😏

erran-r7:
@bcall-r7 during testing I realised that we don't log an error if a stack group did not exist (we just carry on and converge the ones that do).

bin/convection
```diff
@@ -15,9 +15,11 @@ module Convection
     end
 
     desc 'converge STACK', 'Converge your cloud'
+    option :stack_group, :type => :string, :desc => 'The name of a stack group defined in your cloudfile to converge'
+    option :stack_list, :type => :string, :desc => 'A ordered comma separated list of stacks to converge'
```
erran-r7: @jmanero-r7 @bcall-r7 @onefrankguy @jmanero-r7 @athompson-r7 what do you guys think about this command line option name? I wonder if `--stacks` makes more sense than `--stack-list`.

Respond with a 👍 to this comment for `--stacks` or a 👎 for `--stack-list`.

erran-r7:
> - [ ] Converge with stack that is undefined errors correctly (prints error and does not converge)

This scenario fails with:

```
ecarey ~/r/p/c/e/getting-started-guide (issue-108-simplify)> bundle exec convection converge --stack-list bad-instance
/Users/ecarey/rapid7/platform-delivery/convection/lib/convection/control/cloud.rb:29:in `filter_deck': undefined method `call' for nil:NilClass (NoMethodError)
        from /Users/ecarey/rapid7/platform-delivery/convection/lib/convection/control/cloud.rb:35:in `converge'
        from /Users/ecarey/rapid7/platform-delivery/convection/bin/convection:22:in `converge'
        from /Users/ecarey/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
        from /Users/ecarey/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
        from /Users/ecarey/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
        from /Users/ecarey/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
        from /Users/ecarey/rapid7/platform-delivery/convection/bin/convection:74:in `<top (required)>'
        from /Users/ecarey/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/bin/convection:23:in `load'
        from /Users/ecarey/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/bin/convection:23:in `<main>'
```

erran-r7:
> Converge with stack that is undefined errors correctly (prints error and does not converge)

This scenario silently fails:

```
$ bundle exec convection converge bad-stack-name
    converge  Stack vpc1
    complete  Stack vpc1 has no changes
    converge  Stack vpc2
    complete  Stack vpc2 has no changes
    converge  Stack instance1
    complete  Stack instance1 has no changes
    converge  Stack instance2
    complete  Stack instance2 has no changes
    converge  Stack instance3
    complete  Stack instance3 has no changes
```

bin/convection
bturner-r7: :+1: for `--stacks`

lib/convection/control/cloud.rb
```diff
@@ -16,17 +16,23 @@ def stacks
...
       end
 
-      def converge(to_stack, &block)
-        unless to_stack.nil? || stacks.include?(to_stack)
-          block.call(Model::Event.new(:error, "Stack #{ to_stack } is not defined", :error)) if block
-          return
-        end
+      def filter_deck(options = {}, &block)
+        included_stacks = included_stacks(options)
+        return @cloudfile.deck if included_stacks.empty?
```
bturner-r7: This reads weird:  "No included stacks, OK, I'll return all stacks".  Also there a are a few error conditions that aren't caught:
* user specifies a non-existent stack group -> converges all stacks
* user specifies stack group and stack list -> converges stack group

Would something like this work:

```ruby
# throw an error if the user specifies both a stack group and list of stacks
if options[:stack_group] && options[:stack_list]
  block.call(Model::Event.new(:error, "Cannot specify --stack-group and --stack-list at the same time", :error)) if block
  return []
end

# throw an error if the user specifies a nonexistent stack groups
if (options[:stack_group] && !stack_groups.key?(options[:stack_group]))
  block.call(Model::Event.new(:error, "Unknown stack group: #{options[:stack_group]}", :error)) if block
  return []
end

# throw an error if the user specifies nonexistent stacks
if Array(options[:stack_list]).any? { |name| !@cloudfile.stacks.key?(name) }
  bad_stack_names = options[:stack_list].reject { |name| @cloudfile.stacks.key?(name) }
  block.call(Model::Event.new(:error, "Stack#{'s' if bad_stack_names.length > 1} #{bad_stack_names.join(', ')} #{bad_stack_names.length > 1 ? 'do' : 'does'} not exist", :error)) if block
  return []
end

filter = Array(stack_groups[options[:stack_group]] || options[:stack_list])

# if no filter is specified, return the entire deck
return @cloudfile.deck if filter.empty?

@cloudfile.stacks.select { |name, stack| filter.include?(name) }
```

bturner-r7:
Looks like I should read comments before file diffs ;)  @erran-r7 had already caught my comments.

It'd be nice to see some documentation on how to use this new feature.  Otherwise, 👍

erran-r7:
> @erran-r7 had already called out the error conditions I commented on.

You solved it nicely though. 😉

lib/convection/control/cloud.rb
```diff
@@ -44,13 +66,14 @@ def converge(to_stack, &block)
...
-      def diff(to_stack, &block)
-        @cloudfile.deck.each do |stack|
-          block.call(Model::Event.new(:compare, "Compare local state of stack #{ stack.name } (#{ stack.cloud_name }) with remote template", :info))
+      def diff(to_stack, options = {}, &block)
+        filter_deck(options, &block).each do |stack|
+          #pp stack
+          block.call(Model::Event.new(:compare, "Compare local state of stack #{ stack[1].name } (#{ stack[1].cloud_name }) with remote template", :info))
 
-          difference = stack.diff
+          difference = stack[1].diff
```
bturner-r7: Rather than using `[1]`, how about adding `do |stack_name, stack|` to the block parameters above?

lib/convection/control/cloud.rb
```diff
@@ -16,17 +16,40 @@ def stacks
...
+        if options[:stack_group] && !stack_groups.key?(options[:stack_group])
+          block.call(Model::Event.new(:error, "Unknown stack group: #{options[:stack_group]}", :error)) if block
+          return []
+        end
+
+        # throw an error if the user specifies nonexistent stacks
+        if Array(options[:stacks]).any? { |name| !@cloudfile.stacks.key?(name) }
+          bad_stack_names = options[:stacks].reject { |name| @cloudfile.stacks.key?(name) }
+          block.call(Model::Event.new(:error, "Stack#{'s' if bad_stack_names.length > 1}
+                    #{bad_stack_names.join(', ')} #{bad_stack_names.length > 1 ? 'do' : 'does'} not exist", :error)) if block
```
erran-r7: Shall we create a private helper to generate this string? Truncating it like this is a bit confusing, at least for me, at first.

lib/convection/control/cloud.rb
```diff
@@ -16,17 +16,40 @@ def stacks
...
+
+        filter = Array(stack_groups[options[:stack_group]] || options[:stacks])
+
+        # if no filter is specified, return the entire deck
+        return stacks if filter.empty?
+        @cloudfile.stacks.select { |name, _stack| filter.include?(name) }
+      end
+
+      def converge(to_stack, options = {}, &block)
+        filter_deck(options).each do |stack|
```
erran-r7: We now use *both* of the following:

```ruby
filter_deck(options).each do |stack|
```

```ruby
filter_deck(options).each do |stack, stack_name|
```

They can't both be right.

* If we're returning an `Array` *of key/value pairs* — we need to use `... do |(stack_name, stack)|`.
* If we're returning an `Hash` — we need to use `... do |stack_name, stack|`.
* If we're returning a stack — we need to use `... do |stack|`.
* If we're returning a stack name — we need to use `... do |stack_name|`.

bcall-r7: fixed

lib/convection/control/cloud.rb
```diff
@@ -16,17 +16,39 @@ def stacks
...
+        # throw an error if the user specifies a nonexistent stack groups
+        if options[:stack_group] && !stack_groups.key?(options[:stack_group])
+          block.call(Model::Event.new(:error, "Unknown stack group: #{options[:stack_group]}", :error)) if block
+          return []
+        end
+
+        # throw an error if the user specifies nonexistent stacks
+        if Array(options[:stacks]).any? { |name| !@cloudfile.stacks.key?(name) }
+          bad_stack_names = options[:stacks].reject { |name| @cloudfile.stacks.key?(name) }
+          block.call(Model::Event.new(:error, "Non Existent Stack(s) #{bad_stack_names.join(', ')}", :error)) if block
```
erran-r7: Should we be using "Undefined" instead of "Non Existent"? If a stack is defined in your Cloudfile but has not been converged yet then the create call to CloudFormation wouldn't have been called meaning there is no matching stack in cloud formation.

lib/convection/control/cloud.rb
```diff
@@ -68,8 +68,8 @@ def converge(to_stack, options = {}, &block)
       end
 
       def diff(to_stack, options = {}, &block)
-        filter_deck(options, &block).each do |stack_name, stack|
-          block.call(Model::Event.new(:compare, "Compare local state of stack #{ stack_name } (#{ stack.cloud_name }) with remote template", :info))
+        filter_deck(options, &block).each do |_stack_name, stack|
+          block.call(Model::Event.new(:compare, "Compare local state of stack #{ stack.name } (#{ stack.cloud_name }) with remote template", :info))
```
erran-r7: @bcall-r7 wait if we're going to use `stack.name` anyways why even return the entire hash (vs. just the `Stack` values)?

PR #170, opened 2016-08-02
Fix #99 - Support extending the convection resource DSL

### LINKED ISSUE #99: Add native support for extending the convection DSL
This would mean supporting non-cloudformation centric DSL methods. This can be done using mix-ins

One example is that internal to Rapid7 we have a tool which wraps convection to have a VPC-centric DSL. The wrapper added methods for spinning up services which can consistent of multiple components such as a web server and an associated RDS instance with the correct IAM policies setup with it accordingly.

Currently we have to monkey patch certain methods in convection to achieve this goal of adding non-cloudformation centric resource wrappers.

erran-r7:
This pull request adds support for extending the existing convection resources. It gives the ability to create "services" or "resource collections" that can be called inline like the built-in AWS resource methods.

For example a "custom resource"/"resource collection" could be — a "web service". You could pull in this web service helper method to define a set of AWS resources without manually defining each of them. They are configurable using convection attributes and can be implemented using the resource collection DSL or defining a new subclass of the resource collection interface that calls convection's `attach_to_dsl` helper method.

# Examples
## Subclass of the resource collection interface
```ruby
# ./resources/web_service.rb
class WebService < Convection::Model::Template::ResourceGroup
  attach_to_dsl(:web_service) # Defines the `web_service` helper method for use in convection templates

  attribute :public_site # defines a getter/setter named "WebService#public_site"
  attribute :domain # defines a getter/setter named "WebService#domain"
  attribute :subdomain # defines a getter/setter named "WebService#subdomain"

  # The #execute method is executed when rendering convection resources.
  # Overriding it will allow you to define custom resources dynamically based on attributes in your resource collection.
  def execute
    _self = self # you must preserve scope for nested calls.

    ec2_instance "ExampleOrgInstance" do
      # ...

      tag 'Name', _self.name
      tag 'DomainName', _self.domain
      tag 'Fqdn', _self.subdomain ? "#{_self.subdomain}.#{_self.domain}" : "#{_self.domain}"

      endpoint_output_key = _self.public_site ? 'PublicDnsName' : 'PrivateDnsName'
      endpoint_output_value = get_att(_self.name, endpoint_output_key)
      with_output("#{_self.name}Endpoint", endpoint_output_value)
    end

    elb 'ExampleOrgLoadBalancer' do
      # ...
    end if public_site
  end
end
```

```ruby
# ./templates/example_org_service.rb
require 'convection'

EXAMPLE_ORG_SERVICE  = Convection.template do
  web_service 'ExampleDotOrgSite' do
    domain 'example.org'
    subdomain 'www'
    public_site true
  end
end
```

```ruby
# Cloudfile
require_relative './resources/web_service.rb'
require_relative './templates/example_org_service.rb'

stack 'example-org', EXAMPLE_ORG_SERVICE
```

---

# TODO
- [ ] Provide a better summary of this "customisable" convection resource DSL provides to open source users.
- [ ] Provide guidance on how you can pull in these custom resources.
- [ ] ...
- [ ] Remove this list of TODOs once this issue is updated with them all complete.

# Goals
- [x] Support defining a "custom resource" or a collection of resources by subclassing a new interface.
  - [x] This interface should have an abstract implementation which defines some basic things like how to be rendered within a Convection template (into CloudFormation resource JSON).
- [ ] Provide user guides on how to define a collection of resources using library code.
- [x] Internally (to my own team) migrate away from the resource collection system we defined (into convection proper's implementation).

### DIFF
--- a/lib/convection/model/template.rb
+++ b/lib/convection/model/template.rb
@@ -24,6 +24,7 @@ module Convection
               resources[rname] = resource
             end
           end
+          alias attach_resource_group attach_resource
         end
       end
 
--- a/lib/convection/model/template.rb
+++ b/lib/convection/model/template.rb
@@ -27,10 +27,7 @@ module Convection
 
           def attach_resource_group(name, klass)
             define_method(name) do |rname, &block|
-              resource_group = klass.new(rname, self)
-              resource_group.instance_exec(&block) if block
-
-              resource_groups[rname] = resource_group
+              resource_groups[rname] = klass.new(rname, self, &block)
             end
           end
         end
@@ -86,10 +83,7 @@ module Convection
       end
 
       def resource_group(name, &block)
-        r = Model::Template::ResourceGroup.new(name, self)
-
-        r.instance_exec(&block) if block
-        resource_groups[name] = r
+        resource_groups[name] = Model::Template::ResourceGroup.new(name, self, &block)
       end
 
       def output(name, &block)
@@ -235,7 +229,7 @@ module Convection
 
       def execute
         instance_exec(&@definition)
-        resource_groups.each(&:execute)
+        resource_groups.each { |_, group| group.execute }
       end
 
       def render(stack_ = nil)
@@ -250,10 +244,16 @@ module Convection
           'Parameters' => parameters.map(&:render),
           'Mappings' => mappings.map(&:render),
           'Conditions' => conditions.map(&:render),
-          'Resources' => resources.map(&:render),
+          'Resources' => all_resources.map(&:render),
           'Outputs' => outputs.map(&:render),
           'Metadata' => metadata.map(&:render)
         }
+      end
+
+      def all_resources
+        resource_groups.reduce(resources) do |result, (_name, resource_group)|
+          result.merge(resource_group.resources)
+        end
       end
 
       def diff(other, stack_ = nil)
--- a/lib/convection/model/template.rb
+++ b/lib/convection/model/template.rb
@@ -240,12 +240,7 @@ module Convection
           'Parameters' => parameters.map(&:render),
           'Mappings' => mappings.map(&:render),
           'Conditions' => conditions.map(&:render),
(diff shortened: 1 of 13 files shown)


erran-r7:
Marked as do not merge since we'll want to release #169 as a `minor` (`MAJOR.MINOR.PATCH`) release first.

lib/convection/model/template.rb
```diff
@@ -76,6 +82,10 @@ def resource(name, &block)
         resources[name] = r
       end
 
+      def resource_collection(name, &block)
+        resource_collections[name] = Model::Template::ResourceCollection.new(name, self, &block)
+      end
```
bturner-r7: What does this method do?  Same question for the "not yet implemented" method below?

lib/convection/model/template/resource_collection.rb
```diff
@@ -0,0 +1,52 @@
...
+        def execute
+        end
+
+        def run_definition
+          instance_exec(&@definition) if @definition
+        end
+
+        def resource_collection(*)
+          fail NotImplementedError, "#{self.class}#resource_collection is not yet implemented."
+        end
```
bturner-r7: What is this method suppose to do once implemented?

lib/convection/model/template.rb
erran-r7: > What use case do you see for this?

@bturner-r7 I'm glad you spotted this and I meant to bring it up as a "concern". I wasn't really sure on whether we should expose `resource_collection` as a DSL method because like you said:

> Why would you want to define resources inside an inline ResourceCollection as opposed to directly in a template?

I haven't been able to answer this question. We can easily take this out and remove the not implemented error since the nested use-case wouldn't need to be prevented.

erran-r7:
Removing the **do not merge** label since these changes are now ready to be merged down. I'll address the comment around no need for a generic `resource_collection` method and then merge down.

PR #178, opened 2016-08-16
Support deleting stacks using `convection delete`

erran-r7:
Previously we did not expose the `Stack#delete` command in our CLI. Adds a `delete` subcommand that functions like the `diff`/`converge` commands.


bcall-r7:
Verification Steps
- [ ] Delete a stack that does not exist. It should error and not delete anything
- [ ] Delete a specific stack with --stacks
- [ ] Delete a stack_group
- [ ] Delete everything

bcall-r7:
Ship it

PR #180, opened 2016-08-23
Threading status

bcall-r7:
Resolve stack status information in parallel and add a back off to counter rate limiting.

1. Instead if resolving the stack status when doing the instance eval and creating the a new stack class we now resolve the statuses after creating a stack class for each stack defined in the cloud file. Stacks will still converge in the same order as before.

2. A intelligent sleep was added when resolving stack information or diffs fail because AWS rate limiting. This back off works by repeating the request after a sleep that grows by .25 seconds per failed request. This will allow us to remove the sleeps after each diff greatly reducing diff time.

Verification Steps
- [x] Stack delete functions correctly.
- [ ] Diffs in all internal environments show no changes.(expected)
- [x] Stacks with stack.get converge  (i.e correctly resolve stack.get when resolving a resource from a existing stack)
- [x]  Diffs work appropriately. (i.e correctly resolve stack.get when resolving a resource from a existing stack)
- [x] Rate limiting does not fail the program, instead we back off and retry.


bcall-r7:
@jmanero-r7 @athompson-r7 @fmitchell-r7 Opinions?

athompson-r7:
Does this make any order guarantee? i.e. stack1 finishes before stack3 starts

I'm not very clear on the expected outcome here; can we add some docs, comments, etc.

bcall-r7:
@athompson-r7 I have update the PR.

lib/convection/control/stack.rb
```diff
@@ -178,6 +182,11 @@ def status_caller(count = 0.25)
         status_caller(count)
       end
 
+      # Calls The below methods in parallel
+      #   Convection::Model::Stack#get_resource
```
athompson-r7: `get_resources`?

bcall-r7: Good catch. Fixing now

lib/convection/control/stack.rb
```diff
@@ -170,6 +162,42 @@ def cloud_name
...
+      # Calls The below methods in parallel
+      #   Convection::Model::Stack#get_resources
+      #   Convection::Model::Stack#get_template
+      #   Convection::Model::Stack#resource_attributes
+      #   Convection::Model::Stack#get_events
+      def resolver
+        hash = { :get_resources => nil, :get_template => nil, :resource_attributes => nil, :get_events => 1 }
+        threads = []
+        hash.each do |call_me, value|
+          threads << Thread.new do
```
bturner-r7: I think `get_resources` and `resource_attributes` need to be run series and can't be run in parallel because `get_resources` populates the `@resources` instance variable that `resource_attributes` iterates through.  I'd suggest auditing the code to see if there are other similar dependencies, as well as if any semaphores are needed.

bturner-r7:
As discussed on the phone, I'm in favor of using AWS' [built-in retry logic](https://github.com/aws/aws-sdk-ruby/blob/ef9f2d3655c4bacb3adff580a6b295df510f78fe/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb#L16) instead of coding our own.

It might also make sense to split back-off and threading into different PRs to make it easier to review.

PR #174, opened 2016-08-09
Retry expired credentials

erran-r7:
@bcall-r7 @bturner-r7 @eciramella-r7 Thoughts? I extended some changes that I found lying around from the brilliant @athompson-r7. Recalled those when I hit the credential error today.

PR #182, opened 2016-09-09
Call shared config fresh before retry

erran-r7:
cc @bcall-r7 @athompson-r7 @bturner-r7 @eciramella-r7 

See also #174 which started off this work. After some testing we realised that a stack created before a refresh would have stale EC2/CloudFormation clients. We now use the `Aws.shared_config.fresh` function to refresh credentials and re-assign new clients (initialised with the appropriate credentials) to the ec2/cf client instance variables.

### DIFF
--- a/lib/convection/control/stack.rb
+++ b/lib/convection/control/stack.rb
@@ -121,12 +121,8 @@ module Convection
         @attributes = options.delete(:attributes) { |_| Model::Attributes.new }
         @options = options
 
-        client_options = {}.tap do |opt|
-          opt[:region] = @region
-          opt[:credentials] = @credentials unless @credentials.nil?
-        end
-        @ec2_client = Aws::EC2::Client.new(client_options)
-        @cf_client = Aws::CloudFormation::Client.new(client_options)
+        reload_cf_client
+        reload_ec2_client
 
         ## Remote state
         @exist = false
@@ -368,7 +364,7 @@ module Convection
         warn 'AWS Credentials have expired, please re-authenticate...'
         sleep @credential_error_wait_time_seconds
         @credential_error_count += 1
-        Aws.shared_config.fresh
+        reload_cf_client
         retry
       rescue Aws::Errors::ServiceError => e
         @errors << e
@@ -403,7 +399,7 @@ module Convection
         warn 'AWS Credentials have expired, please re-authenticate...'
         sleep @credential_error_wait_time_seconds
         @credential_error_count += 1
-        Aws.shared_config.fresh
+        reload_cf_client
         retry
       rescue Aws::Errors::ServiceError => e
         @errors << e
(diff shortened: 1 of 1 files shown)


lib/convection/control/stack.rb
```diff
@@ -586,12 +582,39 @@ def get_events(pages = nil, stack_name = id)
...
+
+      def client_options
+        client_options = { region: @region }
+        client_options[:credentials] = @credentials unless @credentials.nil?
 
+        client_options
+      end
+
+      def reload_cf_client
+        Aws.shared_config.fresh if @credentials.nil?
```
athompson-r7: Here be dragons...  `Aws.shared_config`[1] not part of the public api.

[1] http://www.rubydoc.info/gems/aws-sdk-core/2.5.1/Aws.shared_config

erran-r7: @athompson-r7 nice catch. I looked more into using public APIs for doing this and, as far as I can tell, the only way to reload the credentials (not the actual file) would be by using [`Aws::AssumeRoleCredentials`](https://github.com/aws/aws-sdk-ruby/blob/ef9f2d3655c4bacb3adff580a6b295df510f78fe/aws-sdk-core/lib/aws-sdk-core/assume_role_credentials.rb) unfortunately. 🙍

This could be something we configured but we'd need a set of persistent credentials that allowed us to the use `Amazon::STS::Client` API to refresh credentials when we're nearing expiry.

That may sound familiar since we do the same in Awsaml and the work in this PR as well as #174 was to enable reloading the credentials Awsaml re-wrote to the credentials file. 😥

athompson-r7: 1. This needs to continue to work with IAM user credentials, have you verified that?
2. Why do we need to refresh the credentials?  Can we create a new Client obj?  Can we let the user refresh the credentials themselves?

It seems like credentials expiring in the middle of the command shouldn't affect the outcome.  That is the command should be immutable.

erran-r7: @athompson-r7 sorry, I was on holiday. Back now!

> This needs to continue to work with IAM user credentials, have you verified that?

I've implemented for compatibility (e.g. an early return if we have set `@credentials`), but I haven't done testing locally.

> Can we create a new Client obj?

Yes, creating a new client object loads credentials from `Aws.shared_config`.

> Why do we need to refresh the credentials?

Otherwise the credentials read at the first client's initialisation will be used.

> Can we let the user refresh the credentials themselves?

It depends:
* Consumers of convection (as library code) **can** refresh credentials themselves.
* Users of the convection binary **must** re-run any commands that were interrupted between requests due to expired credentials.

> It seems like credentials expiring in the middle of the command shouldn't affect the outcome.

The goal of this pull request is to ensure that the request convection sends uses credentials from `~/.aws/credentials` (granted the `:credentials` option was not passed in during initialisation in which case we maintain the previous functionality).

> That is the command should be immutable.

The command itself was previously being interrupted by expired credentials in the middle of multiple operations. This just ensures the latest content of the credentials file is used.

erran-r7:
Closing this after reverting #174 (in #193). This feature requires more work and wasn't deemed necessary as part of the `v1.0` release.

PR #183, opened 2016-09-09
Use retry limit

bcall-r7:
This should speed up diffs by about 2 seconds per stack
Use the retry logic provided by the aws sdk. See here https://github.com/aws/aws-sdk-ruby/blob/ef9f2d3655c4bacb3adff580a6b295df510f78fe/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb#L16 and https://github.com/aws/aws-sdk-ruby/blob/a3a442e0824318dcd95021a1f6057309518bf2a5/aws-sdk-core/spec/aws/s3/client_spec.rb#L15

A custom retry limit can be set in the cloudfile like below or it defaults to 6 which is a ~19 second sleep
```
retry_limit YOUR_NUMBER
```

Verification Steps
- [x] Diffs are functioning properly
- [x] Diffs are not being rate limited

### DIFF
--- a/lib/convection/model/cloudfile.rb
+++ b/lib/convection/model/cloudfile.rb
@@ -14,6 +14,7 @@ module Convection
       attribute :name
       attribute :region
       attribute :splay
+      attribute :retry_limit
 
       ## Helper to define a template in-line
       def template(*args, &block)
@@ -31,7 +32,7 @@ module Convection
         options[:region] ||= region
         options[:cloud] = name
         options[:attributes] = attributes
-
+        options[:retry_limit] = retry_limit
         @stacks[stack_name] = Control::Stack.new(stack_name, template, options, &block)
         @deck << @stacks[stack_name]
       end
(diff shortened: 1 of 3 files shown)


lib/convection/model/cloudfile.rb
```diff
@@ -14,6 +14,7 @@ module Cloudfile
       attribute :name
       attribute :region
       attribute :splay
```
athompson-r7: Don't need to expose this to the Cloudfile anymore right?  Especially since we are removing the sleep logic from  cloud.rb.

athompson-r7:
I like this!

I think this like should be removed too:
https://github.com/rapid7/convection/pull/183/files#diff-42dac5767fb03565ec5b57e48794f6efL86

lib/convection/model/cloudfile.rb
bcall-r7: We wont need it after we remove it from converge/delete. I am only doing these changes for diff as the testing effort is quite substantial. I will do converges once finished with diffs.

athompson-r7:
Looks good to me @bcall-r7!  Thanks for removing my hack ✨

asebastian-r7:
🚢 it!

PR #184, opened 2016-09-13
When doing a diff print errors found during stack initialization

bcall-r7:
When a instance of each stack is created it resolves a bunch of information about that stack. If the request fails because of rate limiting we do not collect the needed info. Then when doing a diff we may assume the stack does not exist or certain attributes are not set and display them as changes. This change will collect errors encountered and display them before discontinuing the diff.
Example
```
$ convection diff
       error  Error(s) during stack diff Rate exceeded
```

Verification Steps
- [x] Run diff where we are rate limited and see that the diff does not happen and a error is returned
- [x] Run a diff not rate limited and see that the diff is done.
- [x] Inject another error and see that both errors are printed.

### DIFF
--- a/lib/convection/control/cloud.rb
+++ b/lib/convection/control/cloud.rb
@@ -114,6 +114,18 @@ module Convection
           return
         end
 
+        errors = []
+        stacks.each_value do |stack|
+          if stack.error?
+            errors << stack.errors.collect { |x| x.exception.message }
+          end
+        end
+        errors = errors.uniq.flatten!
+        unless errors.empty?
+          block.call(Model::Event.new(:error, "Error(s) during stack diff #{errors.join(', ')}", :error), errors) if block
+          return
+        end
+
         filter_deck(options, &block).each_value do |stack|
           block.call(Model::Event.new(:compare, "Compare local state of stack #{ stack.name } (#{ stack.cloud_name }) with remote template", :info))
 


athompson-r7:
LGTM

lib/convection/control/cloud.rb
```diff
@@ -114,6 +114,18 @@ def diff(to_stack, options = {}, &block)
...
+        stacks.each_value do |stack|
+          if stack.error?
+            errors << stack.errors.collect { |x| x.exception.message }
+          end
+        end
+        errors = errors.uniq.flatten!
+        unless errors.empty?
+          block.call(Model::Event.new(:error, "Error(s) during stack diff #{errors.join(', ')}", :error), errors) if block
+          return
+        end
```
bturner-r7: 1. Should this be within `filter_deck` below so that we only throw errors on the current stack-group or filtered stacks?
2. Should we do one `block.call` per stack to more closely follow how we handle errors in the `converge` method?
3. Should we proceed with the diff if an error occurs?  Do we exit with a non-zero exit status when errors occur?

bcall-r7: 1/2. No, if we took that approach we would be doing number 2 where we only error on stacks that have errors and proceed with the diff for stacks in a good state. I do not feel like that is a good idea as there is a possibility the user will miss the error. I feel we should error if any stacks had trouble being initialized. 
3. We should not proceed with the diff if an error occurs. We do not exit with a non zero exit code. (same as other error handling for converges or deletes)

bturner-r7: Should we change the converge behavior to match the diff behavior?  If we report errors on all stacks when only asking to diff a specific stack, we should report errors on all stacks when asking to converge a specific stack.  Any reason for the inconsistency.

Let's open a bug to fix the exit status.  CLI tools should return non-zero exit codes on failures.

lib/convection/control/cloud.rb
```diff
@@ -114,6 +114,18 @@ def diff(to_stack, options = {}, &block)
           return
         end
 
+        errors = []
+        stacks.each_value do |stack|
+          if stack.error?
+            errors << stack.errors.collect { |x| x.exception.message }
```
bturner-r7: When errors happen in converges, we pass back an array of `Error` objects.  For diff errors, we pass back an array of error messages.  Why this inconsistency?\

lib/convection/control/cloud.rb
bcall-r7: Yes we should add that at some point. I see two types of errors here, one during stack initialization, one during the diff/converge operation. For diff we are covered if there is a error during stack init, for converges i think the opposite. I am focusing on fixing the specific issue in diffs at the minute. We will open issues, track the appropriate changes we should make for converges however they will be prioritized after the diff work is complete.

lib/convection/control/cloud.rb
bcall-r7: This is because the different method of error handling. With converges we pass back the error object for the error encountered during the converge. For diffs we are looking for any errors that have already happened and bail if there were any. I guess to say, if the diff caused the error I would take the same approach as converge but as it has already happened we can just collect all the errors and return them. I plan on taking the same approach as i did with diffs to fix converges when checking for initialization errors.

lib/convection/control/cloud.rb
bcall-r7: https://github.com/rapid7/convection/issues/185
https://github.com/rapid7/convection/issues/186

PR #187, opened 2016-09-13
Parallel initialisation

bcall-r7:
# Description
Resolve stack information in parallel. This should cut the init time down almost in half. The number of threads can be set in the cloudfile with the thread_count option. The default of 2 seems to provide the best gain for the least amount of rate limiting. 

# Verification Steps
- [x] Diffs function properly
- [x] Diffs still fail when an error is encountered in init.

### DIFF
--- a/lib/convection/model/cloudfile.rb
+++ b/lib/convection/model/cloudfile.rb
@@ -68,17 +68,17 @@ module Convection
         instance_eval(IO.read(cloudfile), cloudfile, 1)
 
         work_q = Queue.new
-        @deck.each { |stack| work_q.push stack }
+        @deck.each { |stack| work_q.push(stack) }
         workers = (0...@thread_count).map do
           Thread.new do
             until work_q.empty?
               stack = work_q.pop(true)
               stack.status
-              stack.template_info if stack.exist?
+              stack.load_template_info if stack.exist?
             end
           end
         end
-        workers.map(&:join)
+        workers.each(&:join)
       end
     end
   end
--- a/lib/convection/model/cloudfile.rb
+++ b/lib/convection/model/cloudfile.rb
@@ -69,6 +69,7 @@ module Convection
         @deck = []
         @stack_groups = {}
         @thread_count ||= 2
+
         instance_eval(IO.read(cloudfile), cloudfile, 1)
         work_q = Queue.new
         @deck.each { |stack| work_q.push stack }
--- a/lib/convection/model/cloudfile.rb
+++ b/lib/convection/model/cloudfile.rb
@@ -2,6 +2,8 @@ require_relative '../control/stack'
 require_relative '../dsl/helpers'
 require_relative '../model/attributes'
 require_relative '../model/template'
+require 'Benchmark'
+require 'Thread'
 
 module Convection
   module DSL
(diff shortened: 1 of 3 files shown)


lib/convection/model/cloudfile.rb
```diff
@@ -2,6 +2,8 @@
 require_relative '../dsl/helpers'
 require_relative '../model/attributes'
 require_relative '../model/template'
+require 'Benchmark'
```
asebastian-r7: Are we still using this? I saw in your terminal that you were doing "Diff took x milliseconds", but that logic does not appear to be present anymore.

asebastian-r7:
LGTM except for the (possible) unnecessary require of benchmark . :shipit:

lib/convection/model/cloudfile.rb
bcall-r7: nope ill remove thanks

lib/convection/model/cloudfile.rb
```diff
@@ -41,6 +43,10 @@ def stack(stack_name, template, options = {}, &block)
       def stack_group(group_name, group_list)
         @stack_groups[group_name] = group_list
       end
+
+      def init_threads(count)
+        @thread_count = count
+      end
```
athompson-r7: Is this used anywhere, I cannot seem to find a reference.

bcall-r7: Lines 70,75

lib/convection/model/cloudfile.rb
```diff
@@ -61,8 +63,22 @@ def initialize(cloudfile)
         @stacks = {}
         @deck = []
         @stack_groups = {}
+        @thread_count = thread_count || 2
```
erran-r7: Can't you just `@thread_count ||= 2` here?

lib/convection/model/cloudfile.rb
```diff
@@ -61,8 +63,22 @@ def initialize(cloudfile)
...
 
         instance_eval(IO.read(cloudfile), cloudfile, 1)
+
+        work_q = Queue.new
+        @deck.each { |stack| work_q.push stack }
```
erran-r7: **style**: Parenthesis around `stack`: `work_q.push(stack)`.

lib/convection/model/cloudfile.rb
```diff
@@ -61,8 +63,22 @@ def initialize(cloudfile)
...
+        workers = (0...@thread_count).map do
+          Thread.new do
+            until work_q.empty?
+              stack = work_q.pop(true)
+              stack.status
+              stack.template_info if stack.exist?
```
erran-r7: Thoughts on calling this `Stack#load_template_info`?

lib/convection/model/cloudfile.rb
```diff
@@ -61,8 +63,22 @@ def initialize(cloudfile)
...
+            end
+          end
+        end
+        workers.map(&:join)
```
erran-r7: Should we use `Enumerable#each` here since we don't care about the resulting array?

lib/convection/model/cloudfile.rb
bcall-r7: No because the thread_count is the value specified in the cloud file so i only want it to be 2 if its not specified.

PR #192, opened 2016-09-22
Fix #113 - Define Stack#fetch which fails fast when a value is unset

### LINKED ISSUE #113: Fail fast on converge if stack.get returns null
This could also be implemented by creating a `#fetch(stack, key, default_value = nil)` method which raises an exception if no default value (defaulting to `nil` should be allowed) was provided similar to Hash. `#get` can be reimplemented as a call to this method.

Let's get some discussion going on what the desired behaviour would be:

@bturner-r7 added this to the list of general improvements so maybe he could expand on why he would like to fail fast with the current `#get` method. cc @athompson-r7 @jmanero-r7 @simonirwin-r7 @abegley-r7

erran-r7:
Fix #113.


athompson-r7:
@erran Is there a test that validates this new behavior?

erran-r7:
> @erran Is there a test that validates this new behavior?

@athompson-r7 Now there are (since https://github.com/rapid7/convection/pull/192/commits/e0b7517936c3c5f9207f5e15bec35f251908f459). 😜

spec/convection/model/attributes_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      it 'raises a key error if the key was not defined' do
+        expect { subject.fetch('<stack-name>', '<attribute-key>') }.to raise_error(KeyError)
+      end
+
+      ['<truthy object>', true, false].each do |default|
+        it "supports #{default.inspect} as the default value when no value was set" do
+          expect { subject.fetch('<stack-name>', '<attribute-key>', default) }.to_not raise_error
+        end
+
+
```
erran-r7: Extra space. 😡

erran-r7: Resolved by 0126a14.

PR #197, opened 2016-09-27
Fix #196 - Support cross stack references

### LINKED ISSUE #196: Support cross-stack references
# Description
CloudFormation now supports cross stack references. Convection should provide a mechanism to export/import the outputs/references.

See also the CloudFormation [cross-stack reference walkthrough](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/walkthrough-crossstackref.html).

erran-r7:
# Description
Fix #196 - Add support for the new `Export` `Outputs` property and the `Fn::ImportValue` function.

* See also the [Fn::ImportValue](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-importvalue.html) documentation.
* See also the [Export section of the Outputs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/outputs-section-structure.html) documentation.

# Motivation and Context
To simplify cross-stack references written in Convection templates by leveraging the new addition to CloudFormation.

### DIFF
--- a/lib/convection/model/template/resource.rb
+++ b/lib/convection/model/template/resource.rb
@@ -337,11 +337,10 @@ module Convection
           }
         end
 
-        def with_output(output_name = name, value = reference, export_as: nil, &block)
+        def with_output(output_name = name, value = reference, &block)
           o = Model::Template::Output.new(output_name, @template)
           o.value = value
           o.description = "Resource #{ type }/#{ name }"
-          o.export_as = export_as if export_as
 
           o.instance_exec(&block) if block
           @template.outputs[output_name] = o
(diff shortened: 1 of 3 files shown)


lib/convection/model/template/resource.rb
```diff
@@ -337,10 +337,11 @@ def reference
           }
         end
 
-        def with_output(output_name = name, value = reference, &block)
+        def with_output(output_name = name, value = reference, export_as: nil, &block)
```
erran-r7: Is there any reason to have this named parameter? The user can just add the output in the existing configuration block parameter.

simonirwin-r7:
This would have made CR7Services a lot easier, compared to polluting the Output section of each stack.

lib/convection/model/template/resource.rb
erran-r7: Resolved in 7219c24.

PR #199, opened 2016-09-29
Fix #198 - Warn users of manually creating predefined resources

### LINKED ISSUE #198: Warn users when using the Template#resource method for an already supported resource type
**Feature request**: Users should be warned when they define a resource inline if a resource of the same type exists.

## Observed behaviour
Calling `resource` from within a template does not display a warning.

## Desired behavior
**Feature request**: Users should be warned when they define a resource inline if a resource of the same type exists.

erran-r7:
# Description
Fix #198 - Warn users of when they redefine resources.

### DIFF
--- a/spec/convection/model/template/validate_bytesize_spec.rb
+++ b/spec/convection/model/template/validate_bytesize_spec.rb
@@ -21,13 +21,13 @@ class Convection::Model::Template
           description 'Validations Test Template - Excessive Bytesize'
 
           200.times do |count|
-            resource "EC2_INSTANCE_#{count}" do
-              type 'AWS::EC2::Instance'
-              property 'AvailabilityZone', 'us-east-1a'
+            ec2_instance "EC2_INSTANCE_#{count}" do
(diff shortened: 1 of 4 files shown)


spec/convection/model/template/validate_bytesize_spec.rb
```diff
@@ -21,13 +21,13 @@ class Convection::Model::Template
           description 'Validations Test Template - Excessive Bytesize'
 
           200.times do |count|
-            resource "EC2_INSTANCE_#{count}" do
-              type 'AWS::EC2::Instance'
-              property 'AvailabilityZone', 'us-east-1a'
+            ec2_instance "EC2_INSTANCE_#{count}" do
+              availability_zone 'us-east-1a'
               property 'ImageId', 'ami-76e27e1e'
               property 'KeyName', 'test'
```
erran-r7: These can be replaced with the existing methods vs. re-defining the properties.

**side note**: Should we warn when re-defining properties as well?

PR #200, opened 2016-09-29
Fix accidental overloading of status breaking converges

bcall-r7:
# Description
Fix accidental overloading of status. This causes issues in converge as in_progress? needs status and calls the incorrect one(mine).

# Motivation and Context
Converges are broken
# Testing Steps
Converges work again

### DIFF
--- a/lib/convection/model/cloudfile.rb
+++ b/lib/convection/model/cloudfile.rb
@@ -73,7 +73,7 @@ module Convection
           Thread.new do
             until work_q.empty?
               stack = work_q.pop(true)
-              stack.status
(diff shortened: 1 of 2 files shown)


lib/convection/model/cloudfile.rb
```diff
@@ -73,7 +73,7 @@ def initialize(cloudfile)
           Thread.new do
             until work_q.empty?
               stack = work_q.pop(true)
-              stack.status
+              stack.template_status
```
erran-r7: Thoughts on changing this to `Stack#refresh!`?

PR #201, opened 2016-09-29
Fix #189 - Introduce a more explicit error around expired credentials

### LINKED ISSUE #189: Update messaging to represent the real issue when "converging" fails
In this particular case the user's AWS credentials expired, so they were unable to continue polling the events for the stack that was being created.  This is fine as Cloudformation is doing the right thing in the background, but the error presented to the user is confusing suggesting that something failed and that they need to do something about it.

We need to differentiate between failures in fetching events and converging.

```
create_in_progress  StagingCDNBucketPolicy: (AWS::S3::BucketPolicy/)
       error  Error converging stack s3-buckets
* The security token included in the request is expired
```

erran-r7:
# Description
Fix #189.

Add `Stack#credential_error?` for testing whether all the errors present
are credential errors (if there are other errors handle regularly).

### DIFF
--- a/lib/convection/control/cloud.rb
+++ b/lib/convection/control/cloud.rb
@@ -75,7 +75,10 @@ module Convection
           block.call(Model::Event.new(:converge, "Stack #{ stack.name }", :info)) if block
           stack.apply(&block)
 
-          if stack.error?
+          if stack.credential_error?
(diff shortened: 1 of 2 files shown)


bcall-r7:
Waiting on merging this until we finish issue #203

PR #202, opened 2016-09-29
Fix #185 - Use a nonzero exit code when failing to diff/converge

### LINKED ISSUE #185: CLI errors should exit non zero
Diff or converge errors do not exit with a nonzero exit and they should.

erran-r7:
# Description
Fix #185 — Use `Kernel#exit` instead of early `return`s or `break`s to ensure a non-zero exit code.


bcall-r7:
This is going to change after issue 203 is merged so lets hold off on this.

PR #214, opened 2016-10-13
Add support for ELB v2 Application Load Balancer resources/properties

bcall-r7:


lib/convection/model/template/resource/aws_elb.rb
```diff
@@ -25,6 +25,7 @@ class ELB < Resource
           property :policy, 'Policies', :type => :list
           property :scheme, 'Scheme'
           property :security_group, 'SecurityGroups', :type => :list
+          alias_method :security_groups, :security_group
```
erran-r7: 😅

lib/convection/model/template/resource/aws_elbv2_listener.rb
```diff
@@ -0,0 +1,31 @@
...
+  module Model
+    class Template
+      class Resource
+        ##
+        # AWS::ElasticLoadBalancingV2::Listener
+        ##
+        class ELBV2Listener < Resource
+          type 'AWS::ElasticLoadBalancingV2::Listener', :elbv2_listener
+          property :certificates, 'Certificates', :type => :list
+          property :default_actions, 'DefaultActions', :type => :list
```
erran-r7: Use the singular here for consistency.

lib/convection/model/template/resource/aws_elbv2_listener.rb
```diff
@@ -0,0 +1,31 @@
...
+module Convection
+  module Model
+    class Template
+      class Resource
+        ##
+        # AWS::ElasticLoadBalancingV2::Listener
+        ##
+        class ELBV2Listener < Resource
+          type 'AWS::ElasticLoadBalancingV2::Listener', :elbv2_listener
+          property :certificates, 'Certificates', :type => :list
```
erran-r7: >Use the singular here for consistency.

lib/convection/model/template/resource/aws_elbv2_listener.rb
```diff
@@ -0,0 +1,31 @@
...
+          property :default_actions, 'DefaultActions', :type => :list
+          property :load_balancer_arn, 'LoadBalancerArn'
+          property :port, 'Port'
+          property :protocol, 'Protocol'
+          property :ssl_policy, 'SslPolicy'
+
+          # FIXME: handle certificates
```
erran-r7: 😏

lib/convection/model/template/resource/aws_elbv2_listener_rule.rb
```diff
@@ -0,0 +1,35 @@
...
+  module Model
+    class Template
+      class Resource
+        ##
+        # AWS::ElasticLoadBalancingV2::ListenerRule
+        ##
+        class ELBV2ListenerRule < Resource
+          type 'AWS::ElasticLoadBalancingV2::ListenerRule', :elbv2_listener_rule
+          property :actions, 'Actions', :type => :list
+          property :conditions, 'Conditions', :type => :list
```
erran-r7: >Use the singular here for consistency.

lib/convection/model/template/resource/aws_elbv2_load_balancer.rb
```diff
@@ -0,0 +1,38 @@
...
+        ##
+        class ELBV2LoadBalancer < Resource
+          include Model::Mixin::Taggable
+
+          type 'AWS::ElasticLoadBalancingV2::LoadBalancer', :elbv2_load_balancer
+          property :load_balancer_attributes, 'LoadBalancerAttributes', :type => :list
+          property :name, 'Name'
+          property :scheme, 'Scheme'
+          property :security_group, 'SecurityGroups', :type => :list
+          alias_method :security_groups, :security_group
```
erran-r7: Let's just keep these methods singular and move to usage of the singular nouns.

lib/convection/model/template/resource/aws_elbv2_target_group.rb
```diff
@@ -0,0 +1,54 @@
...
+          property :health_check_protocol, 'HealthCheckProtocol'
+          property :health_check_timeout_seconds, 'HealthCheckTimeoutSeconds'
+          property :healthy_threshold_count, 'HealthyThresholdCount'
+          property :matcher, 'Matcher', :type => :hash
+          property :name, 'Name'
+          property :port, 'Port'
+          property :protocol, 'Protocol'
+          property :target_group_attributes, 'TargetGroupAttributes', :type => :list
+          property :targets, 'Targets', :type => :list
+          alias_method :target_descriptions, :targets
```
erran-r7: Why "target descriptions"?

lib/convection/model/template/resource/aws_elbv2_target_group.rb
```diff
@@ -0,0 +1,54 @@
...
+          include Model::Mixin::Taggable
+
+          type 'AWS::ElasticLoadBalancingV2::TargetGroup', :elbv2_target_group
+          property :health_check_interval_seconds, 'HealthCheckIntervalSeconds'
+          property :health_check_path, 'HealthCheckPath'
+          property :health_check_port, 'HealthCheckPort'
+          property :health_check_protocol, 'HealthCheckProtocol'
+          property :health_check_timeout_seconds, 'HealthCheckTimeoutSeconds'
+          property :healthy_threshold_count, 'HealthyThresholdCount'
+          property :matcher, 'Matcher', :type => :hash
```
erran-r7: Is this matcher actually a simple JSON object or should it be a resource property?

lib/convection/model/template/resource/aws_elbv2_listener.rb
rdickey-r7: Fixed in a later commit.  The singular is now a function, which is consistent with code elsewhere.

lib/convection/model/template/resource/aws_elbv2_listener.rb
rdickey-r7: Fixed in a later commit.  The singular is now a function, which is consistent with code elsewhere.

lib/convection/model/template/resource/aws_elbv2_target_group.rb
rdickey-r7: it's a stupid simple json object (only one possible key) but I made it a resource_property anyway, for consistency

lib/convection/model/template/resource/aws_elbv2_target_group.rb
rdickey-r7: See:  http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-targetgroup.html

The json for the TargetGroup type says "Targets", but the actual object type is the "TargetDescription":
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-targetgroup-targetdescription.html

lib/convection/model/template/resource/aws_elbv2_listener.rb
rdickey-r7: fixed in a later commit

lib/convection/model/template/resource/aws_elbv2_listener_rule.rb
rdickey-r7: singular was made a function in a later commit.

lib/convection/model/template/resource/aws_elbv2_load_balancer.rb
rdickey-r7: I'm curious why we want to be inconsistent with what amazon defined with cloudformation?  Looking at the docs:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-loadbalancer.html
They define it as "SecurityGroups" because it's a list not a single thing, which makes sense.  We could then add a singular function to add a single security group, which would make sense but might be a breaking change.

lib/convection/model/template/resource/aws_route53_recordset.rb
```diff
@@ -9,20 +9,34 @@ class Resource
         ##
         class Route53RecordSet < Resource
           type 'AWS::Route53::RecordSet', :route53_recordset
-          property :alias_target, 'AliasTarget'
+          property :alias_tgt, 'AliasTarget'
           property :comment, 'Comment'
           property :failover, 'Failover'
-          property :geo_location, 'GeoLocation', :array
+          property :geo_loc, 'GeoLocation'
```
rdickey-r7: @erran-r7 See alias_target comment above

lib/convection/model/template/resource/aws_elbv2_listener_rule.rb
```diff
@@ -0,0 +1,35 @@
...
+            action.instance_exec(&block) if block
+            actions << action
+          end
+
+          # Append a condition
+          def rule_condition(&block)
+            condition = ResourceProperty::ELBV2ListenerRuleCondition.new(self)
+            condition.instance_exec(&block) if block
+            conditions << condition
+          end
```
bturner-r7: Similar to above, thoughts on using a `condition` alias or renaming the method to `condition` to be more consistent with singular/plural patterns?

lib/convection/model/template/resource/aws_elbv2_listener.rb
```diff
@@ -0,0 +1,35 @@
...
+          property :load_balancer_arn, 'LoadBalancerArn'
+          property :port, 'Port'
+          property :protocol, 'Protocol'
+          property :ssl_policy, 'SslPolicy'
+
+          def listener_certificate(&block)
+            certificate = ResourceProperty::ELBV2ListenerCertificates.new(self)
+            certificate.instance_exec(&block) if block
+            certificates << certificate
+          end
```
rdickey-r7: I just re-checked the AWS docs:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listener.html

and I think you're correct that it should just be named "certificate".  Sometimes, Amazon will use one thing as the json key, and the value is a different type (such as Targets vs TargetDescription at http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-targetgroup.html) but that's not the case here.  Good catch.

lib/convection/model/template/resource/aws_elbv2_listener_rule.rb
rdickey-r7: Same as above, I will change it.

erran-r7: @bturner-r7 @rdickey *beware of dragons* eek, @bcall-r7 and I discovered that having overriding `condition` will cause issues rendering the CF JSON (there is a top level `condition` method that is override if you re-define a property of that name). That condition method is used to fill the `Conditions` CF field.

@bcall-r7 did I summarise that right? I can't recall exactly what scenario you exposed that issue under.

lib/convection/model/template/resource/aws_route53_recordset.rb
```diff
@@ -9,20 +9,34 @@ class Resource
         ##
         class Route53RecordSet < Resource
           type 'AWS::Route53::RecordSet', :route53_recordset
-          property :alias_target, 'AliasTarget'
+          property :alias_tgt, 'AliasTarget'
```
erran-r7: > but I could also make it backwards-comparable.

Since we're bumping the version by a major version I don't mind breaking this so much. We just shouldn't release `1.0.0` until we've cleared up all these discrepancies.

We can search for our internal convection usage to see if either of these properties were used and update to use your format before releasing a new pre-release version. Can you go ahead and create a ticket in our internal issue tracker for that (CC @bcall-r7 and I)?

lib/convection/model/template/resource/aws_elbv2_listener_rule.rb
erran-r7: @rdickey I think it's grand. Can you add an `# @note ...` yardoc comment explaining it above?

rdickey-r7:
@erran-r7 @bturner-r7 I think your concerns have been addressed.

PR #216, opened 2016-11-01
Add S3 Website Configuration and properties

rdickey-r7:
# Description
Add ability to specify a website_configuration to an s3 bucket.  To do correctly, this required creating a number of resource_property types to support the various options that CloudFormation gives.

# Motivation and Context
This change is necessary to enable static website hosting for an s3 bucket.

# Usage Examples

    s3_bucket 'bucketName' do
      ...
      website_configuration do
        index_document 'index.html'
        redirect_all_requests_to do
          host_name 'somehost.example.com'
          protocol 'https'
        end
      end
    end

# Testing Steps

# Post-merge Steps

### DIFF
--- a/lib/convection/model/template/resource_property/aws_cloudfront_defaultcachebehavior.rb
+++ b/lib/convection/model/template/resource_property/aws_cloudfront_defaultcachebehavior.rb
@@ -12,6 +12,7 @@ module Convection
           property :compress, 'Compress'
           property :default_ttl, 'DefaultTTL'
           property :forwarded_values, 'ForwardedValues'
+          property :max_ttl, 'MaxTTL'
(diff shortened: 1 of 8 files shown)


lib/convection/model/template/resource_property/aws_cloudfront_defaultcachebehavior.rb
```diff
@@ -12,6 +12,7 @@ class CloudFrontDefaultCacheBehavior < ResourceProperty
           property :compress, 'Compress'
           property :default_ttl, 'DefaultTTL'
           property :forwarded_values, 'ForwardedValues'
+          property :max_ttl, 'MaxTTL'
```
erran-r7: Unrelated, but safe to include.

PR #217, opened 2016-11-02
Maintain backwards compatability with route53

rdickey-r7:
# Description

Changes alias_target and geo_location in a route53_recordset to allow either json or a resource_property

# Motivation and Context

Previously, users needed to specify the json for alias_target and
geo_location for a route53_recordset.  In the add-alb branch, merged in
PR 214, this was changed in a backwards-incompatible way, requiring
anyone who had previously specified json to now specify a block for the
resource_property.  Now, either way is acceptable.

This is potentially a paradigm that can be used elsewhere, as there are
many places where a resource_property would be appropriate, but has not
existed and users have implemented the functionality directly with a
json (or ruby hash) object.  This will allow us to deprecate that method
in favor of the more clean resource_property approach, without breaking
everything in a 1.0 upgrade.

# Usage Examples

# Testing Steps

# Post-merge Steps

### DIFF
--- a/lib/convection/model/template/resource/aws_route53_recordset.rb
+++ b/lib/convection/model/template/resource/aws_route53_recordset.rb
@@ -26,16 +26,28 @@ module Convection
           property :record_type, 'Type'
           property :weight, 'Weight'
 
+          # Add new resource_property functionality
           def alias_target(&block)
             a = ResourceProperty::Route53AliasTarget.new(self)
             a.instance_exec(&block) if block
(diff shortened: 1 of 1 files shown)


rdickey-r7:
I am going to change this to have a single method that handles both a block and a resource_property arg.  I misunderstood the results of my testing.
http://stackoverflow.com/questions/9373104/why-doesnt-ruby-support-method-overloading

PR #219, opened 2016-11-08
Show when resource will be replaced during a convection diff

rdickey-r7:
# Description

This change affects the "convection diff" output when Amazon wants to "update" a resource using CloudFormation in a way that requires [Replacement](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-replacement).

# Motivation and Context

The problem with updates requiring replacement is that you have to "just know" what property changes require an update.  Now, when a qualifying property change will produce a create/destroy action, the diff will show "replace" in red.

# Usage Examples

This example, from this repo, shows how changing a VPC Cidr will cause a converge to replace the VPC:
screen shot 2016-11-08 at 4 11 53 pm

# Testing Steps

Before converging any changes you would like to make (or, after making a few that require replacement and a few that require updates just to see what happens) run "convection diff".  The --vv shown in the example is optional


lib/convection/model/replace_properties.rb
```diff
@@ -0,0 +1,399 @@
...
+  module Model
+    # Unfortunately, there are a lot of resources to include in this array,
+    # and rather than parsing out only the ones that we need right now
+    # and potentially missing it in the future, just disable rubocop whining
+    # about that particular error:
+    # rubocop:disable ClassLength
+    class Diff
+      # Properties for which a change requires a replacement
+      # (as opposed to an in-place update)
+      # Create/update this list with tmp/process_html.sh
```
erran-r7: ❤️ ❤️ ❤️ ❤️

lib/convection/model/replace_properties.rb
```diff
@@ -0,0 +1,399 @@
+module Convection
+  module Model
+    # Unfortunately, there are a lot of resources to include in this array,
+    # and rather than parsing out only the ones that we need right now
+    # and potentially missing it in the future, just disable rubocop whining
+    # about that particular error:
```
erran-r7: 👍

lib/convection/model/diff.rb
```diff
@@ -12,23 +13,32 @@ class Diff
...
       def initialize(key, ours, theirs)
         @key = key
         @ours = ours
         @theirs = theirs
 
-        @action = if ours && theirs then :update
-                  elsif ours then :create
-                  else :delete
+        @action = if ours && theirs
+                    property_name = key[/AWS::[A-Za-z0-9:]+\.[A-Za-z0-9]+/]
```
erran-r7: Really nice use of a regex index here. I love when you find a good use case for this.

Issue #203, opened 2016-09-30
Support converging multiple clouds at the same time


erran-r7:
# Description
Users should be able to pass multiple Cloudfiles into commands such as `diff` and `converge`.

PR #220, opened 2016-11-09
Support setting the default --delayed-output behaviour with CONVECTION

erran-r7:
# Description
## Before
You must specify the `--delayed-output` flag as `true` to get rid of the warning.

## After
* Setting `--delayed-output` explicitly using the flag now silences the warning regardless of whether it was set to true/false.
* Setting `CONVECTION_DELAY_OUTPUT` explicitly using the flag now silences the warning regardless of whether it was set to true/false.

---

This shows the user has not set the environment variable and has not passed in the command line flag.

@bcall-r7 @rdickey


erran-r7:
@bcall-r7 can you give this a quick review?

bin/convection
```diff
@@ -105,9 +105,10 @@ module Convection
       def operation(task_name, stack)
         work_q = Queue.new
         semaphore = Mutex.new
-        unless options[:delayed_output]
-          puts 'For easier reading when using multiple cloudfiles output can be delayed until task completion.'
-          puts 'If you would like delayed output please use the "--delayed_output true" option.'
+        if options[:delayed_output].nil?
```
erran-r7: Should we add a condition that `cloudfiles.length` is `>= 2` to print this?

rdickey-r7:
I gave this a shot on antisocial-convection and got this error:

```

{master* rdickey@aus-mbp-3074:~/rapid7/antisocial-convection}$ bundle install|grep convec
Using convection 1.0.0.pre.beta.666 from source at `~/rapid7/convection`
Using convection-r7-services 0.3.22
Using convection-r7-spinnaker 0.1.7
{master* rdickey@aus-mbp-3074:~/rapid7/antisocial-convection}$ convection diff --vv
/Users/rdickey/rapid7/convection/bin/convection:22:in `<class:CLI>': undefined method `default_delay_output' for Convection:Module (NoMethodError)
	from /Users/rdickey/rapid7/convection/bin/convection:9:in `<module:Convection>'
	from /Users/rdickey/rapid7/convection/bin/convection:5:in `<top (required)>'
	from /Users/rdickey/.rvm/gems/ruby-2.3.3@antisocial-convection/bin/convection:22:in `load'
	from /Users/rdickey/.rvm/gems/ruby-2.3.3@antisocial-convection/bin/convection:22:in `<main>'
	from /Users/rdickey/.rvm/gems/ruby-2.3.3@antisocial-convection/bin/ruby_executable_hooks:15:in `eval'
	from /Users/rdickey/.rvm/gems/ruby-2.3.3@antisocial-convection/bin/ruby_executable_hooks:15:in `<main>'
{master* rdickey@aus-mbp-3074:~/rapid7/antisocial-convection}$
```

Did I miss something?

rdickey-r7:
As a more general comment on the intent of this change, if I were to modify the behavior of the "delayed output" warning, I would have removed the warning completely, and simply made a note in the help and readme.  It is missing from both.

There is no value in having this feature advertised so prominently, and doing so is a distraction from the output which I want to see.  Generally speaking, "warnings" of this type should only be used to announce deprecations, where later versions may take away existing functionality.  This does not appear to be the case for delayed vs undelayed output, but do correct me if I'm wrong.

erran-r7:
> I gave this a shot on an internal convection stack repo and got this error:

> Did I miss something?

No, looks like we don't load the whole `convection` codebase only a subset. https://github.com/rapid7/convection/blob/71c7323acb/bin/convection#L3 We'll need to `require 'convection'` at the top of the binary to get this to work. I haven't ran from this branch in ages, so I can't recall why this didn't fail for me in my testing.

erran-r7:
> As a more general comment on the intent of this change, if I were to modify the behavior of the "delayed output" warning, I would have removed the warning completely, and simply made a note in the help and readme. It is missing from both.
> 
> There is no value in having this feature advertised so prominently, and doing so is a distraction from the output which I want to see. Generally speaking, "warnings" of this type should only be used to announce deprecations, where later versions may take away existing functionality. This does not appear to be the case for delayed vs undelayed output, but do correct me if I'm wrong.

@rdickey-r7 I agree there. This was definitely intended to be "backwards compatible" by making it obvious to the user. I'd be happy documenting the feature as you stated. I know @bcall-r7 did a lot of thought into the original implementation in #203 (before we release the `1.x` code line).

bcall-r7:
Yup @erran-r7 is correct. It was displayed as such to make it obvious to the users after the transition. Whatever happens to it now is fine with me. I agree its not really needed.

rdickey-r7:
Cool sounds good, and understood about wanting to make it obvious at first.  Maybe what I said can be implemented at a later date.

PR #222, opened 2016-11-11
Support EC2 Spot Fleet resources

abegley-r7:
# Description
Add CloudFormation resources and resource properties to support EC2 Spot Fleet. 

An internal pull request depends on these changes.


lib/convection/model/template/resource_property/aws_ec2_iam_instance_profile.rb
```diff
@@ -0,0 +1,15 @@
+require_relative '../resource_property'
```
simonirwin-r7: This filename looks wrong?

erran-r7: Nope, this looks right. It will load the original `Convection::Model::Template::ResourceProperty` before appending to its namespace.

PR #224, opened 2016-12-20
Fix #223 - Do not cast errors when diffing

### LINKED ISSUE #223: Undefined method error #message for String
# Description
When the AWS API request expires a undefined method error is raised.

```
us-east-1        error  Error(s) during stack initialization Request has expired., The security token included in the request is expired, The security token included in the request is expired
/Users/ecarey/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/convection-1.0.0.pre.beta.9/bin/convection:200:in block in emit_events': undefined methodmessage' for "Request has expired.":String (NoMethodError)
```

erran-r7:
Fix #223 — get rid of the undefined method error by removing cast to string.

Delete and converges stopped this when we updated error handling logic.

### DIFF
--- a/lib/convection/control/cloud.rb
+++ b/lib/convection/control/cloud.rb
@@ -129,9 +129,7 @@ module Convection
           # Find errors during diff
           emit_credential_error_and_exit!(stack, &block) if stack.credential_error?
           if stack.error?
-            errors = stack.errors.collect { |x| x.exception.message }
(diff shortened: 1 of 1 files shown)


erran-r7:
This doesn't change what I was trying to change on double taking it. It only removes the string message. The errors array was passed successfully before.

PR #225, opened 2016-12-22
Add target_group_arns to asg resource

rdickey-r7:
# Description
Support the `TargetGroupARNs` resource property.


cdoughty-r7:
👍

PR #230, opened 2017-01-13
Fix #229 -- Return a list a availability zones

### LINKED ISSUE #229: describe_availability_zones test double return list of regions rather than list of availability_zones
# Description
The [describe_availability_zones](https://github.com/rapid7/convection/blob/master/spec/ec2_client_context.rb#L3) double actually returns a list of regions

## Expected behavior
I would expect this to return a list of availablity_zones like: `eu-central-1a, eu-central-1b`

## Observed behavior
This actually returns a list of regions like this: `eu-central-1, eu-west-1`

## Steps to Reproduce

You can see this behavior in [this](https://github.com/rapid7/convection/blob/failing-describe-availability-zones-spec/spec/convection/control/stack_spec.rb#L34-L39) failing spec.

athompson-r7:
This changes the `describe_availability_zones` double so that it returns
a list of availability_zones.

Fixes #229

### DIFF
--- a/spec/convection/control/stack/after_create_tasks_spec.rb
+++ b/spec/convection/control/stack/after_create_tasks_spec.rb
@@ -38,7 +38,7 @@ module Convection::Control
       it 'is executed on Stack#apply' do
         subject.apply
 
-        expect(task.availability_zones).to include('eu-central-1')
+        expect(task.availability_zones).to include(a_string_starting_with('eu-central-1'))
       end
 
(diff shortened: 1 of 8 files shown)


athompson-r7:
@erran-r7/@bcall-r7 can you guys take a look at this?

athompson-r7:
Thanks @erran-r7

erran-r7:
No problem!


Thanks, Erran
________________________________

PR #231, opened 2017-01-17
Add describe-subnets subcommand to dump subnets by AZ

simonirwin-r7:
# Description
This adds a `describe-subnets` subcommand.

Usage is:
```
Usage:
  convection describe-subnets

Options:
  [--cloudfile=CLOUDFILE]  
                           # Default: Cloudfile
  [--format=FORMAT]        
                           # Default: json
                           # Possible values: json, yaml

Describe subnets for the VPC
```

This subcommand will dump all VPC subnets (keyed by AZ) to JSON or YAML.

### DIFF
--- a/bin/convection
+++ b/bin/convection
@@ -188,10 +188,10 @@ module Convection
 
       def describe_vpc_subnets(format)
         vpc_template = @cloud.stacks['vpc'].instance_variable_get(:@current_template)
-        raise "No template defined for [vpc] stack" if vpc_template.nil?
+        raise 'No template defined for [vpc] stack' if vpc_template.nil?
 
         subnet_map = {}
-        vpc_template['Resources'].select { |_name,attrs| attrs['Type'] == 'AWS::EC2::Subnet' }.each do |name, attrs|
+        vpc_template['Resources'].select { |_name, attrs| attrs['Type'] == 'AWS::EC2::Subnet' }.each do |name, attrs|
           subnet_map[attrs['Properties']['AvailabilityZone']] ||= []
           subnet_map[attrs['Properties']['AvailabilityZone']] << { name: name, cidr: attrs['Properties']['CidrBlock'] }
         end
--- a/bin/convection
+++ b/bin/convection
@@ -97,6 +97,14 @@ module Convection
       @cloud.stacks[stack].validate
     end
 
+    desc 'describe-subnets', 'Describe subnets for the VPC'
+    option :cloudfile, :type => :string, :default => 'Cloudfile'
+    option :format, :type => :string, :default => 'json', :enum => %w(json yaml)
+    def describe_subnets
+      init_cloud
+      describe_vpc_subnets(options[:format])
+    end
+
     no_commands do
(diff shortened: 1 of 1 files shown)


bin/convection
```diff
@@ -178,6 +186,23 @@ module Convection
...
       end
 
+      def describe_vpc_subnets(format)
+        vpc_template = @cloud.stacks['vpc'].instance_variable_get(:@current_template)
+        raise 'No template defined for [vpc] stack' if vpc_template.nil?
+
+        subnet_map = {}
+        vpc_template['Resources'].select { |_name, attrs| attrs['Type'] == 'AWS::EC2::Subnet' }.each do |name, attrs|
+          subnet_map[attrs['Properties']['AvailabilityZone']] ||= []
+          subnet_map[attrs['Properties']['AvailabilityZone']] << { name: name, cidr: attrs['Properties']['CidrBlock'] }
```
erran-r7: 1. I think `:subnet_id` be very useful.
2. I'd prefer seeing a structure that lists the subnets, if we call this `describe-subnets` I wouldn't expect availability zone keys (would expect that as a property of each subnet).


I imagined a format similar to the following:

```json
{
  "subnets": [
    {
      "name": "TestSubnet0",
      "subnet_id": "subnet-deadb33f",
      "availability_zone": "us-east-1a",
      "cidr": "10.0.0.0/27"
    },
    {
      "name": "TestSubnet1",
      "subnet_id": "subnet-d34db33f",
      "availability_zone": "us-east-1e",
      "cidr": "10.0.0.192/27"
    }
  ]
}
```

bin/convection
```diff
@@ -178,6 +186,23 @@ module Convection
...
+        end
+
+        if format == 'json'
+          puts JSON.pretty_generate subnet_map
```
erran-r7:
```ruby
# a bit confusing - had to re-read this. What is the input here and where is the receiver?
puts JSON.pretty_generate subnet_map

# puts is clearly the receiver method and the right hand side is the input
puts JSON.pretty_generate(subnet_map)
```

bin/convection
```diff
@@ -178,6 +186,23 @@ module Convection
...
+        vpc_template = @cloud.stacks['vpc'].instance_variable_get(:@current_template)
+        raise 'No template defined for [vpc] stack' if vpc_template.nil?
+
+        subnet_map = {}
+        vpc_template['Resources'].select { |_name, attrs| attrs['Type'] == 'AWS::EC2::Subnet' }.each do |name, attrs|
+          subnet_map[attrs['Properties']['AvailabilityZone']] ||= []
+          subnet_map[attrs['Properties']['AvailabilityZone']] << { name: name, cidr: attrs['Properties']['CidrBlock'] }
+        end
+
+        if format == 'json'
```
erran-r7: 1. Could've used `case`/`when` instead:

```ruby
case format
when 'json'
when 'yaml'
else
  # default response (argument error)? — though right now this method only has one caller which does validation of format.
end
```

PR #232, opened 2017-01-18
Add ability to exclude Array of AvailabilityZones

### LINKED ISSUE #228: Provide method for modifying the list of availability_zones
# Description
There may be cases where AWS is restricting access to one of the available availability_zones or you just want to deploy to a subset of the available availability_zones.

In either case, there should be a method to modify the list of [availability_zones](https://github.com/rapid7/convection/blob/5c6235a103e323898613382e1d8b104706a018ed/lib/convection/control/stack.rb#L429-L439) such that a subset of the availability_zones are returned.

athompson-r7:
# Description

This change adds the ability to remove an Array of AvailabilityZones
from the Array that is returned in
`Convection::Control::Stack#availability_zones`.

Fixes #228

# Motivation and Context

This is necessary in cases where AWS marks an AvailabilityZone as
`available` but it [isn't _available_ for customer use](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-regions-availability-zones).

# Usage Examples

Adding something like the following to the `Cloudfile` will enable the feature.
```
exclude_availability_zones %w(eu-central-1a)
```

# Testing Steps

# Post-merge Steps


spec/convection/control/stack_spec.rb
```diff
@@ -36,6 +36,28 @@ module Convection::Control
         azs.uniq!
         expect(azs.size).to be(1)
       end
+
+      it 'availability_zones can be ignored' do
+        subject.exclude_availability_zones = %w(eu-central-1a)
+        expect(subject.availability_zones).to contain_exactly('eu-central-1b')
```
rdickey-r7: It may be worthwhile to add a comment here that if Amazon adds an eu-central-1c, this test will break and need to be fixed.  That should save a little time/headache if it does happen.

spec/convection/model/cloudfile_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      it 'can get region name' do
+        subject.region 'eu-central-1'
+        expect(subject.region).to eq('eu-central-1')
+      end
+
+      it 'can exclude_availability_zones an availability_zone' do
+        subject.exclude_availability_zones %w(eu-central-1a)
+        expect(subject.exclude_availability_zones).to contain_exactly('eu-central-1a')
```
rdickey-r7: Same issue with possible failure if Amazon expands zones in eu-central-1

.travis.yml
```diff
@@ -7,4 +7,5 @@ rvm:
   - 2.2.0
   - 2.2.1
   - 2.2.2
+  - 2.3.0
```
rdickey-r7: Any reason not to add 2.3.1, 2.3.3, and/or 2.4.0 while you're adding stuff?  I would leave out 2.3.2 because it's broken.

lib/convection/control/stack.rb
```diff
@@ -432,7 +434,15 @@ def watch(poll = 2, &block)
       #   the call to ec2 client's describe availability zones
       def availability_zones(&block)
         @availability_zones ||=
-          @ec2_client.describe_availability_zones.availability_zones.map(&:zone_name).sort
+          @ec2_client.describe_availability_zones.availability_zones.map(&:zone_name)
+
+        unless @exclude_availability_zones.empty?
+          @exclude_availability_zones.each { |az| @availability_zones.delete(az) }
+        end
+        if @availability_zones.empty? && block
```
erran-r7: Why does it matter if `block` is defined here?

lib/convection/control/stack.rb
```diff
@@ -432,7 +434,15 @@ def watch(poll = 2, &block)
       #   the call to ec2 client's describe availability zones
       def availability_zones(&block)
         @availability_zones ||=
-          @ec2_client.describe_availability_zones.availability_zones.map(&:zone_name).sort
+          @ec2_client.describe_availability_zones.availability_zones.map(&:zone_name)
+
+        unless @exclude_availability_zones.empty?
+          @exclude_availability_zones.each { |az| @availability_zones.delete(az) }
```
erran-r7: No functional difference but I'd have expected something like:

```ruby
@availability_zones.reject! { |az| @exclude_availability_zones.include?(az) }
```

spec/convection/control/stack_spec.rb
```diff
@@ -36,6 +36,28 @@ module Convection::Control
...
+
+      it 'multiple availability_zones can be ignored' do
+        subject.exclude_availability_zones = %w(eu-central-1a eu-central-1b)
+        expect(subject.availability_zones).to contain_exactly
+      end
+
+      it 'remove all availability_zones fails' do
+        subject.exclude_availability_zones = %w(eu-central-1a eu-central-1b)
+        b = proc do
+          puts 'hi'
```
erran-r7: No reason to print anything here — it'll also add junk output to the test output.

spec/convection/model/cloudfile_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+          subnet_length 24
+          enable_dns
+        end
+      end
+    end
+
+    describe 'stack' do
+      subject do
+        klass = Class.new
+        klass.extend(Convection::DSL::Cloudfile)
```
erran-r7: `klass = Class.new(Convection::DSL::Cloudfile)` would've been equivalent to the two lines above.

spec/convection/model/cloudfile_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'spec_helper'
+require 'stringio'
+
+module Convection::Model
+  describe Cloudfile do
+    let(:template) do
```
erran-r7: Looks like we don't use this. 🔨

spec/convection/control/stack_spec.rb
erran-r7: We don't actually call out to amazon to list zones. We stub the response so this should be fine based on the stub today. Even if they expand their zones our stub is hit not the API.

spec/convection/model/cloudfile_spec.rb
erran-r7: See my above commend on using a stub for the availability zone lookup.

.travis.yml
athompson-r7: It's the [ruby-version](https://github.com/rapid7/convection/blob/master/.ruby-version) we specify.

lib/convection/control/stack.rb
athompson-r7: It allows me to test removing all of the azs without an exception.  I can remove it and refactor the test :man_shrugging:

spec/convection/model/cloudfile_spec.rb
athompson-r7: `Convection::DSL::Cloudfile` is a `module` so that will not work.

spec/convection/control/stack_spec.rb
rdickey-r7: Ok, then this is fine how it is.

Issue #234, opened 2017-01-19
stack.availability_zones fails if exclude_availability_zones not set in Cloudfile


athompson-r7:
# Description

For some reason [@exclude_availability_zones](https://github.com/rapid7/convection/blob/master/lib/convection/control/stack.rb#L112) if defaulting to `nil` even though we set the default to an empty Array.

This was introduced in #232 

## Expected behavior

`@exclude_availability_zones` should default to an empty Array if it is not specified in the Cloudfile.

## Observed behavior

The following exception is thrown because `@exclude_availability_zones` is defaulting to `nil` not an empty Array.

## Steps to Reproduce

Install version 1.0.4 of the gem and run `convection diff` on a stack that includes and the Cloudfile does **not** include `exclude_availability_zones`:
```
stack.availability_zones do
  # something
end
```

# Stack Trace or Log Messages

<details>
<summary>Stack trace for `brief error context`</summary>
<pre>
.../convection-1.0.4/lib/convection/control/stack.rb:441:in `availability_zones': undefined method `empty?' for nil:NilClass (NoMethodError)
.../templates/ec2/vpc.rb:132:in `block (2 levels) in <module:EC2>'
.../convection-1.0.4/lib/convection/model/template.rb:23:in `instance_exec'
...snip...
</pre>
</details>

erran-r7:
Resolved in #235.

PR #235, opened 2017-01-20
Only pass exclude_availability_zones option if set

athompson-r7:
# Description
This change will only pass the `exclude_availability_zones option` to the `Stack` controller if the option is set in the Cloudfile.

# Motivation and Context

This address #234 which was introduced in #232.

@erran-r7 @rdickey-r7

### DIFF
--- a/lib/convection/model/cloudfile.rb
+++ b/lib/convection/model/cloudfile.rb
@@ -33,7 +33,7 @@ module Convection
       # @see Convection::Control::Stack#initialize
       def stack(stack_name, template, options = {}, &block)
         options[:region] ||= region
-        options[:exclude_availability_zones] = exclude_availability_zones
+        options[:exclude_availability_zones] = exclude_availability_zones unless exclude_availability_zones.nil?
         options[:cloud] = name
         options[:attributes] = attributes
         options[:retry_limit] = retry_limit
--- a/spec/convection/control/stack_spec.rb
+++ b/spec/convection/control/stack_spec.rb
@@ -57,6 +57,10 @@ module Convection::Control
       it 'can get default availability_zones' do
         expect(subject.availability_zones).to contain_exactly('eu-central-1a', 'eu-central-1b')
       end
+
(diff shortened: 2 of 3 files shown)


lib/convection/model/cloudfile.rb
```diff
@@ -33,7 +33,7 @@ def attribute(stack, key, value)
       # @see Convection::Control::Stack#initialize
       def stack(stack_name, template, options = {}, &block)
         options[:region] ||= region
-        options[:exclude_availability_zones] = exclude_availability_zones
+        options[:exclude_availability_zones] = exclude_availability_zones unless exclude_availability_zones.nil?
```
rdickey-r7: Would `.empty?` work better than `.nil?` ?

rdickey-r7:
Ignore previous comment about `nil?`, I was thinking `blank?` which is only in rails.

spec/convection/control/stack_spec.rb
```diff
@@ -57,6 +57,10 @@ module Convection::Control
       it 'can get default availability_zones' do
         expect(subject.availability_zones).to contain_exactly('eu-central-1a', 'eu-central-1b')
       end
+
+      it 'can get default exclude_availability_zones' do
+        expect(subject.exclude_availability_zones).to match_array([])
```
erran-r7: We could also use `expect(...).to be_empty` here.

spec/convection/model/cloudfile_spec.rb
```diff
@@ -3,11 +3,32 @@
...
+      end
+    end
+
     describe 'stack' do
       subject do
-        klass = Class.new
-        klass.extend(Convection::DSL::Cloudfile)
-        klass
+        # initialize does too much to call that directly
+        # setup only what we need for testing... :/
```
erran-r7: Out of scope for this PR but — Would be nice to move a lot of the stuff it doesn't into a method we can stub for unit testing. Doesn't provide any functional benefit it would be more just developer friendliness.

PR #239, opened 2017-02-14
Print out error message correctly when there is one

rdickey-r7:
# Description
Commit a7c43b71c7f93c9250877155cae9840e52c3db4a broke error message printing by trying to print a variable that doesn't exist.  This fixes that glitch.

# Motivation and Context

# Usage Examples

# Testing Steps

# Post-merge Steps

### DIFF
--- a/bin/convection
+++ b/bin/convection
@@ -240,7 +240,7 @@ module Convection
 
         errors.each do |error|
           error = RuntimeError.new(error) if error.is_a?(String)
-          say "* #{ error_message }"
+          say "* #{ error.message }"
           error.backtrace.each { |trace| say "    #{ trace }" }
         end
       end


erran-r7:
@rdickey-r7 Ah crap, think there may be a bigger issue here. I think I changed that call because there was a undefined method error for some cases. :person_frowning:

rdickey-r7:
@erran-r7 Either way, error_message is wrong.  I think you could have solved the problem that you were trying to solve with only the RuntimeError.new line, and not changing error.message to error_message.  So I think this is ok now (at least, it works for me)

erran-r7:
Think this will break one of converge or diff but fix the other (whichever you were having an issue with). I'll check this out now. Sorry got caught up with other stiffness.

erran-r7:
@rdickey-r7 never mind, the commit from myself *with* your changes resolves #227. The casting is done in the case that `error` is a string so `error.message` will give the string error message. I believe there was a bad rebase before merging that PR. Just validated your changes. 👍 🎉 Thanks again!

PR #257, opened 2017-07-25
Lock mixlib-shellout to 2.2.7 for pre ruby 2.2 compatibility

erran-r7:
# Description
This should fix the issue @mbroomfield-r7 is getting unrelated to his changes.

**NOTE**: We may just be better locking to `>= 2.3` and bumping a major version to prevent similar issues from happening in the foreseeable future.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,4 +11,6 @@ group :development do
   gem 'simplecov'
   gem 'thor-scmversion', '= 1.7.0'
   gem 'yard'
+  # TODO: Remove mixlib dependency once we lock to Ruby >= 2.2.
+  gem 'mixlib-shellout', '= 2.2.7'
 end


dgreene-r7:
Why are we preserving 2.1.x compatibility in the first place?

erran-r7:
> Why are we preserving 2.1.x compatibility in the first place?

@dgreene-r7 I have no objections to dropping it. It's in our travis matrix which screwed @mbroomfield-r7 over though.

erran-r7:
@dgreene-r7 I can remove 2.1 from the matrix and add the required ruby version flag in our gem spec then do a major version bump to release Matt's changes?

dgreene-r7:
Sounds good to me!

PR #259, opened 2017-08-02
Feature: Add EFS support

fmitchell-r7:
# Description
This adds support for creating [AWS::EFS::FileSystem](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-efs-filesystem.html) and [AWS::EFS::MountTarget](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-efs-mounttarget.html) resources.

# Motivation and Context
We want to be able to manage [Amazon Elastic File System](https://aws.amazon.com/efs/) resources with Convection.

# Usage Examples
```ruby
# Assign this to EC2 instances that should have access to the EFS
ec2_security_group 'EFSAccess' do
end

ec2_security_group 'EFSMount' do
  ingress_rule(:tcp, 2049, fn_ref('EFSAccess'))
end

efs_file_system 'FileSystem' do
  performance_mode 'maxIO'
  tag 'Owner', 'Alice'
end

efs_mount_target 'MountTarget' do
  file_system_id fn_ref('FileSystem')
  security_groups fn_ref('EFSMount')
  subnet_id stack.get('vpc', 'subnet')
end
```

The [`FileSystemTags`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-efs-filesystem-filesystemtags.html) property on the `AWS::EFS::FileSystem` resource uses the usual `tag` syntax instead of exposing a new `file_system_tag` method.

# Testing Steps
Run `rake` to run the tests and lint checks.


lib/convection/model/template/resource/aws_efs_file_system.rb
```diff
@@ -0,0 +1,24 @@
+require_relative '../resource'
+
+module Convection
+  module Model
+    class Template
+      class Resource
+        # Represents an {https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-efs-filesystem.html
+        # Amazon EFS File System}
+        class EFSFileSystem < Resource
+          include Model::Mixin::Taggable
```
rdickey-r7: Since the actual name is FileSystemTags, I wonder if it might be worth adding an alias for file_system_tag just to keep the naming consistent with what Amazon uses

lib/convection/model/template/resource/aws_efs_mount_target.rb
```diff
@@ -0,0 +1,19 @@
...
+  module Model
+    class Template
+      class Resource
+        # Represents an {https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-efs-mounttarget.html
+        # Amazon EFS Mount Target}
+        class EFSMountTarget < Resource
+          type 'AWS::EFS::MountTarget'
+          property :file_system_id, 'FileSystemId'
+          property :ip_address, 'IpAddress'
+          property :security_groups, 'SecurityGroups', :type => :list
```
rdickey-r7: Should there be a "security_group" helper function, a la https://github.com/rapid7/convection/blob/c60c2e4a22db513a4774a60a7c44c3a59c7e2d2e/lib/convection/model/template/resource/aws_api_gateway_usage_plan.rb#L18 or does that get handled automagically somehow?

rdickey-r7: nevermind, it looks like this is a list of strings not a list of other AWS types, so it should be fine as-is

lib/convection/model/template/resource/aws_efs_file_system.rb
fmitchell-r7: @rdickey-r7, good idea! I added a `file_system_tag` alias in https://github.com/rapid7/convection/pull/259/commits/8fe117c38c90b84137ac26ed0d46da3ca32de971.

PR #262, opened 2017-09-13
Add the AcmCertificateArn property to CloudFront

bturner-r7:
# Description

This adds support for using ACM certificates in CloudFront distributions.

# Motivation and Context
New SSL certificates, even imported 3rd party certificates, use the new ACM certificate functionality in AWS.  This replaces the old IAM certificates.  To use these in convection, the CloudFormation `AcmCertificateArn` property must be used.

# Usage Examples
```ruby
cloudfront_distribution 'FooCloudFront' do
  config do
    ...
    viewer_certificate do
      acm_certificate_arn 'arn:aws:acm:us-east-1:XXXXXXXXXXXX:certificate/XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX'
      ssl_support_method 'sni-only'
    end
  end
end
```

# Testing Steps
Converged a stack using this branch.

### DIFF
--- a/lib/convection/model/template/resource_property/aws_cloudfront_viewercertificate.rb
+++ b/lib/convection/model/template/resource_property/aws_cloudfront_viewercertificate.rb
@@ -9,6 +9,7 @@ module Convection
         class CloudFrontViewerCertificate < ResourceProperty
           property :use_default, 'CloudFrontDefaultCertificate'
           property :iam_certificate, 'IamCertificateId'
+          property :acm_certificate, 'AcmCertificateArn'
           property :minimum_protocol_version, 'MinimumProtocolVersion'
           property :ssl_support_method, 'SslSupportMethod'
         end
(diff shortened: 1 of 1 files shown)


lib/convection/model/template/resource_property/aws_cloudfront_viewercertificate.rb
```diff
@@ -9,6 +9,7 @@ class ResourceProperty
         class CloudFrontViewerCertificate < ResourceProperty
           property :use_default, 'CloudFrontDefaultCertificate'
           property :iam_certificate, 'IamCertificateId'
+          property :acm_certificate, 'AcmCertificateArn'
```
rdickey-r7: According to `egrep -r 'property .+Arns?' lib`, the typical standard is that something like 'AcmCertificateArn' would become acm_certificate_arn.

bturner-r7: makes sense, will update.

bturner-r7: Done

rdickey-r7:
Thanks.  Looks good now.

Issue #256, opened 2017-07-24
Support DeadLetterConfig for Lambda


gschneider-r7:
# Description

Lambda model should support DeadLetterConfig: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-deadletterconfig.html

erran:
It should be noted that this has been possible in convection using the generic `property` setter since well before this ticket was made.

gschneider-r7:
To provide an example of what @erran mentioned:

```ruby
property 'DeadLetterConfig', :TargetArn => 'sqs_or_sns_arn'
```

gschneider-r7:
Available in version 2.2.26

PR #263, opened 2017-09-13
Add more properties to Lambda and make it taggable

gschneider-r7:
# Description

Add KmsKeyArn for environment variable encryption.
Add DeadLetterConfig for failure DeadLetterQueue use. (#256)
Add Taggable mixin to allow tags on lambdas.

# Motivation and Context

I want to use tags and decided to fill in the other missing bits as well.

# Usage Examples

```ruby
lambda_function 'MyLambda' do
  handler 'lambda.handler'
  runtime 'python2.7'
  timeout 30
  memory_size 128
  kms_key_arn get_att('MyKmsKey', 'Arn')

  dead_letter_config do
    target_arn get_att('MyDeadLetterQueue', 'Arn')
  end

  tag 'my_first_tag_key', 'tag_value'
  tag 'my_second_tag_key, 'other_tag_value'
end
```

Of course you can use raw strings as well:
```ruby
kms_key_arn 'arn:aws:kms:us-east-1:XXXXXXXXXXXX:key/my-alias-or-uuid'

dead_letter_config do
  target_arn 'arn:aws:sqs:us-east-1:XXXXXXXXXXXX:my-queue-name'
end
```
# Testing Steps

todo

### DIFF
--- a/lib/convection/model/template/resource/aws_lambda_function.rb
+++ b/lib/convection/model/template/resource/aws_lambda_function.rb
@@ -8,6 +8,8 @@ module Convection
         # AWS::Lambda::Function
         ##
         class Lambda < Resource
+          include Model::Mixin::Taggable
+
           type 'AWS::Lambda::Function'
           property :function_name, 'FunctionName'
           property :description, 'Description'
@@ -16,10 +18,12 @@ module Convection
           property :runtime, 'Runtime'
           property :timeout, 'Timeout'
           property :role, 'Role'
+          property :kms_key_arn, 'KmsKeyArn'
           # psuedo-property definitions. We add the expected name as a nested DSL for these below.
           property :env, 'Environment'
           property :function_code, 'Code'
           property :vpc_cfg, 'VpcConfig'
+          property :dead_letter_cfg, 'DeadLetterConfig'
 
           # Add code block
           def code(&block)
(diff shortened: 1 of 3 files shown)


lib/convection/model/template/resource_property/aws_lambda_dead_letter_config.rb
```diff
@@ -0,0 +1,15 @@
+require_relative '../resource_property'
+
+module Convection
+  module Model
+    class Template
+      class ResourceProperty
+        # Represents an {http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-deadletterconfig.html
+        # Lambda Function DeadLetterConfig}
+        class LambdaDeadLetterConfig < ResourceProperty
```
rdickey-r7: I think the more proper name for this would be LambdaFunctionDeadLetterConfig, and the filename should also reflect that, a la LambdaFunctionCode in the aws_lambda_function_code.rb file.  In contrast, LambdaVpcConfig was done incorrectly based on the standard used in most places of reflecting amazon's naming as closely as possible.

lib/convection/model/template/resource/aws_lambda_function.rb
```diff
@@ -40,6 +44,12 @@ def vpc_config(&block)
             vpc_cfg.instance_exec(&block) if block
             properties['VpcConfig'].set(vpc_cfg)
           end
+
+          def dead_letter_config(&block)
+            env = ResourceProperty::LambdaDeadLetterConfig.new(self)
```
rdickey-r7: This will need to be changed per my other comment in this review.

PR #264, opened 2017-09-13
Fix travis bundler errors

gschneider-r7:
# Description

Force travis to update bundler and system gems as pre-install step.

See [this bug](https://github.com/bundler/bundler/issues/3558) in bundler for context.

Additionally:
- Update travis config to use latest ruby versions
- Update travis config to only build master branch changes (avoids duplicated builds on PRs and releases/tags)
- Update .ruby-version so I don't get prompted when using 2.3.3+ 😄 
# Motivation and Context

Travis-CI builds were failing on 2.2.0 and 2.2.1 versions of Ruby.

# Testing Steps

See if this PR gets a green checkmark!


gschneider-r7:
Should probably update the ruby versions while I'm here, but I'll wait for others to chime in first.

rdickey-r7:
I'd be fine with just testing the latest 2.2.x and 2.3.x, if you wanted to change the ruby versions.  Currently that's 2.2.8 and 2.3.5.  But this is fine as is, so up to you.

gschneider-r7:
I updated the config to use the latest releases for each 2.2, 2.3, and 2.4 versions. Unfortunately RVM's aliases are behind a few releases so the full version had to be specified. This means that it will get out of date over time and need to be updated again manually in the future.

rdickey-r7:
👍

PR #265, opened 2017-10-31
Add '--exclude-stacks' option for 'diff' and 'converge'

rlynch-r7:
# Description
Adds a new option `--exclude-stacks [STACK_NAME] <[STACK_NAME]> ...` to the `convection diff` and `convection converge` operations. Any stacks specified therein (as a space-separated list of stack names) are removed from the `diff` or `converge` operation. If one or more specified stacks does not exist, that's an error.

This new option cannot be used in conjunction with either of the existing options `--stack-group` or `--stacks`.

# Motivation and Context
Increased flexibility while deploying changes to environments.

### DIFF
--- a/lib/convection/control/cloud.rb
+++ b/lib/convection/control/cloud.rb
@@ -22,31 +22,54 @@ module Convection
       end
 
       def filter_deck(options = {}, &block)
-        # throw an error if the user specifies both a stack group and list of stacks
-        if options[:stack_group] && options[:stacks]
-          block.call(Model::Event.new(:error, 'Cannot specify --stack-group and --stack-list at the same time', :error)) if block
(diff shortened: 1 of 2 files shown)


lib/convection/control/cloud.rb
```diff
@@ -22,31 +22,54 @@ def stack_groups
...
+            bad_stack_names = options[:exclude_stacks].reject { |name| @cloudfile.stacks.key?(name) }
+            block.call(Model::Event.new(:error, "Undefined Stack(s) #{bad_stack_names.join(', ')}", :error)) if block
+            return {}
+          end
+
+          filter = Array(options[:exclude_stacks])
+          result = {}
+          stacks.each do |stack_name, stack|
+            unless filter.include? stack_name
+              result[stack_name] = stack
```
bcall-r7: Consider using reject here.

PR #268, opened 2017-11-16
Add support to mark all resources in a stack with the "Retain" Deletion Policy

### LINKED ISSUE #267: Create --retain flag for diff/converge
This command will provide the ability to mark all resources under a stack as Retain unless the deletion policy of it was explicitly set. If any deletion policies were explicitly set this should raise an error with the resource names that caused this. 
 
Some information to help get started with this ticket:
Convection's CLI can be found at bin/convection and uses thor.
Using init_cloud we can populate the @cloud variable giving us access to the @cloud.stacks["<stack-name>"] function.
We can operate on the Stack class from there.
Once can use Stack#template then Template#all_resources to get all resources from the template.
Each Resource object supports the #deletion_policy function which exposes the DeletionPolicy cloud formation attribute.
We should set the deletion policy on each resource inside of the stack's template.
We can copy logic for diff/converge for the dry run and update operations.

mburns-r7:
Fix #267.

- [x] Add a retain flag to `convection diff`
- [x] Add a retain flag to `convection converge`


.rubocop_todo.yml
```diff
@@ -59,7 +59,7 @@ Metrics/AbcSize:
 # Offense count: 34
 # Configuration parameters: CountComments, ExcludedMethods.
 Metrics/BlockLength:
-  Max: 136
+  Max: 140
```
erran-r7: ❤️ 🔨 Thor's`no_commands` block.

lib/convection/control/stack.rb
```diff
@@ -279,8 +278,8 @@ def to_json(pretty = false)
       #   template (in CloudFormation) and the state of the rendered
       #   template (what *would* be converged).
       # @see Convection::Model::Template#diff
-      def diff
-        @template.diff(@current_template)
+      def diff(retain: false)
+        @template.diff(@current_template, retain: retain)
```
erran-r7: Document what the `retain` keyword arg does.

lib/convection/control/stack.rb
```diff
@@ -322,17 +321,17 @@ def resource_dependent_changes?
       #
       # @param block [Proc] a configuration block to pass any
       #   {Convection::Model::Event}s to.
-      def apply(&block)
+      def apply(retain: false, &block)
```
erran-r7: Document the retain kwarg.

lib/convection/model/template.rb
```diff
@@ -263,7 +263,12 @@ def render(stack_ = nil)
           'Parameters' => parameters.map(&:render),
           'Mappings' => mappings.map(&:render),
           'Conditions' => conditions.map(&:render),
-          'Resources' => all_resources.map(&:render),
+          'Resources' => all_resources.map do |resource|
+            if retain && resource.deletion_policy.nil?
+              resource.deletion_policy('Retain')
+            end
+            resource.render
```
erran-r7: 🎨 Add a newline above this to separate the conditional stanza and this return value.

lib/convection/model/template.rb
```diff
@@ -275,12 +280,12 @@ def all_resources
...
 
-      def diff(other, stack_ = nil)
-        render(stack_).diff(other).map { |diff| Diff.new(diff[0], *diff[1]) }
+      def diff(other, stack_ = nil, retain: false)
+        render(stack_, retain: retain).diff(other).map { |diff| Diff.new(diff[0], *diff[1]) }
       end
 
-      def to_json(stack_ = nil, pretty = false)
-        rendered_stack = render(stack_)
+      def to_json(stack_ = nil, pretty = false, retain: false)
```
erran-r7: 📝 Document the new retain kwarg.

erran-r7:
@mburns-r7 We can merge this with or without the docs changes. It'd be nice to have them.

I'll add a bit of context here though:

For users of cloud formation (via convection) the new `--retain` [and `--no-retain`] command line flag(s), including the associated `:retain` boolean keyword, adds support for setting the DeletionPolicy on all resources where it is unset to `"Retain"`.

When setting the Deletion Policy to Retain:

1. On stack update any resources removed will skip deletion.
2. On stack deletion all resources with the Retain deletion policy will skip deletion.

PR #275, opened 2017-12-07
Adds ability to set 'EbsOptimized' and 'Monitoring' properties on EC2Instance

jwark:
# Description
Adds ability to set 'EbsOptimized' and 'Monitoring' properties (both boolean) on EC2Instance, as per CloudFormation docs:
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html#cfn-ec2-instance-ebsoptimized
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html#cfn-ec2-instance-monitoring


erran:
Whoa, only now realising this was opened 12 days ago. Sorry for the delay. Rapid7 has been having holiday parties recently so this was lost amongst that along with new feature support to support exporting to terraform (e.g. #270).

erran:
@jwark @SuperGeno These changes are available in [the latest convection release v2.2.7](https://rubygems.org/gems/convection/versions/2.2.7). Thanks for your patience. 🎉 



---

Also, not sure if you knew this but in the future if you're waiting on this type of change being merged/released properties can always be defined as any object that supports `#to_json` like so:

```ruby
property 'EbsOptimized', true
property 'Monitoring', false
```

jwark:
Thanks! Not a problem at all, we had just monkey-patched the class while waiting for the pull-request to be merged. ;)
Enjoy the rest of your holidays!

PR #278, opened 2017-12-13
Support a hash of additional hcl files to generate

erran-r7:
Add support for using `Resource#additional_hcl_files`/`ResourceCollection#additional_hcl_files` to generate additional HCL files.

Usage:

```ruby
# Values can be any object that generates a JSON Object on `#to_json`.
def additional_hcl_files
  {
    'hello_world.tf.json': { aws_s3_bucket: { mybucketname: { ... } } },
  }
end
```

### DIFF
--- a/bin/convection
+++ b/bin/convection
@@ -244,12 +244,19 @@ module Convection
       private
 
       def import_resources(_resource_name, resource)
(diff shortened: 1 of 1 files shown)


simonirwin-r7:
nice! lgtm

PR #287, opened 2018-01-19
Support aws_s3_bucket terraform export

mburns-r7:
# Description

# Motivation and Context

# Usage Examples

# Testing Steps

# Post-merge Steps

### DIFF
--- a/lib/convection/model/template/resource/aws_s3_bucket.rb
+++ b/lib/convection/model/template/resource/aws_s3_bucket.rb
@@ -50,6 +50,50 @@ module Convection
             properties['WebsiteConfiguration'].set(config)
           end
 
+          def full_name(terraform_import: false)
+            return "#{name.downcase}.#{stack._original_cloud}.#{stack._original_region}.rapid7.com" if terraform_import
+
+            "#{name.downcase}.#{stack.cloud}.#{stack.region}.rapid7.com"
+          end
+
+          def hcl_imports(module_path: 'root')
+            commands = ['# Run the following commands to import your infrastructure into terraform management.', '# ensure :module_path is set correctly', '']
+            module_prefix = module_path.tr('.', '-') if module_path == 'root'
+            result = {}
+
+            commands << '# Import s3 bucket and s3 bucket policy: '
+            commands << "terraform import #{module_prefix}aws_s3_bucket.#{name.downcase} #{full_name(terraform_import: true)}"
+						commands << ''
+						commands
+          end
+
+          def to_hcl_json(module_path: 'root')
+            bucket_resource = {
+              name.downcase => {
+                bucket: full_name,
+                acl: "private",
+                force_destroy: false,
+              }
+            }
+
+            data = [{aws_region: {current: {current: true } } }]
+            tf_resources = [
+              { aws_s3_bucket: bucket_resource },
+            ]
+
+            bucket_reference = "${aws_s3_bucket.#{name.downcase}.id}"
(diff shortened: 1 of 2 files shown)


lib/convection/model/template/resource/aws_s3_bucket.rb
```diff
@@ -50,6 +50,50 @@ def website_configuration(&block)
             properties['WebsiteConfiguration'].set(config)
           end
 
+          def full_name(terraform_import: false)
```
erran-r7: You should grab the physical ID of the resource using `stack.resources[name]` instead of trying to generate it we do like we do in the internal bucket extension which always gets a predictable suffix.

lib/convection/model/template/resource/aws_s3_bucket.rb
```diff
@@ -50,6 +50,50 @@ def website_configuration(&block)
...
+            return "#{name.downcase}.#{stack._original_cloud}.#{stack._original_region}.rapid7.com" if terraform_import
+
+            "#{name.downcase}.#{stack.cloud}.#{stack.region}.rapid7.com"
+          end
+
+          def hcl_imports(module_path: 'root')
```
erran-r7: 1. `hcl_imports` should be `terraform_import_commands` otherwise it will not be called automatically when `terraform-export` is run.

lib/convection/model/template/resource/aws_s3_bucket.rb
```diff
@@ -50,6 +50,50 @@ def website_configuration(&block)
...
+            commands = ['# Run the following commands to import your infrastructure into terraform management.', '# ensure :module_path is set correctly', '']
+            module_prefix = module_path.tr('.', '-') if module_path == 'root'
+            result = {}
+
+            commands << '# Import s3 bucket and s3 bucket policy: '
+            commands << "terraform import #{module_prefix}aws_s3_bucket.#{name.downcase} #{full_name(terraform_import: true)}"
```
erran-r7: Use the physical ID as shown above not an assumed name.

lib/convection/model/template/resource/aws_s3_bucket.rb
```diff
@@ -50,6 +50,50 @@ def website_configuration(&block)
...
+            bucket_reference = "${aws_s3_bucket.#{name.downcase}.id}"
+              policy_json = resources["#{name}Policy"] && resources["#{name}Policy"].document.document.to_json.gsub(full_name, bucket_reference)
+              policy_resource = {
+                name.downcase => {
+                  bucket: bucket_reference,
+                  policy: policy_json.render.to_json,
+                }
+              }
+
+              tf_resources << { aws_s3_bucket_policy: policy_resource }
```
erran-r7: The value returned from this function should be in the format `{ resource: [] }`.

lib/convection/model/template/resource/aws_s3_bucket.rb
```diff
@@ -50,6 +50,50 @@ def website_configuration(&block)
...
+              }
+            }
+
+            data = [{aws_region: {current: {current: true } } }]
+            tf_resources = [
+              { aws_s3_bucket: bucket_resource },
+            ]
+
+            bucket_reference = "${aws_s3_bucket.#{name.downcase}.id}"
+              policy_json = resources["#{name}Policy"] && resources["#{name}Policy"].document.document.to_json.gsub(full_name, bucket_reference)
```
erran-r7: 1. You should generate this HCL in the `aws_s3_bucket_policy` resource class. We can't be sure that open source users (or even other users at work) are placing policies in the same stack as the bucket resource.
2. In the policy resource you can get the bucket name from the bucket attribute.

PR #299, opened 2018-06-05
 Adds support for creating AWS::SNS::Subscription resource and rspec test for it. 

eromani-r7:
Took @jwark 's AWS::SNS::Subscription resource and added unit test plus tested resource creation in the dev account.


jwark:
Thanks @eromani-r7 !  😀

